PR #63, opened 2014-10-22
Refactoring DbUp and providers.

SimonRupar:
Based on #53. Done only changes in providers. 

1. Added methods in DbUp.StandardExtensions for calling scripts from file system in absolute or relative way.

2. Changed logging of exception in UpgradeEngine. Only return exception message, not whole Exception.ToString(). Some fixes regarding logging has been changed also in SqlSciptExecutor.

3. I have created classes IQueryProvider, QueryProviderBase, SqlServerQueryProvider) which contains all SQL statements used in source. Based on my previous experience this can be a life-saver, because you don't need to recheck all source code for SQL statements when something change in databases. All providers extend QueryProviderBase and have their own SQL statements written for some functionality.

4. All providers and their classes are in separate projects. Only exception is MsSql Server, which is still in base library DbUp. I think it would be also useful if classes used for MsSql server is moved to separate project.


SimonRupar:
@JakeGinnivan Because it has been some time since last check-in I am just wondering if there are any opened issues, because I think I finished my part.

SimonRupar:
Closing this pull-request because I cound not do rebase to latest verions.

PR #75, opened 2014-12-09
Refactor MySql support

namzat:
I added some tests and refactored the code to handle multi-line scripts as well as scripts that contain a DELIMITER statement.


JakeGinnivan:
Hey @namzat, I know it has been **ages** but I have finally got around to looking into this. I am opening up a pr at #113 and will get this in soon.

I am a bit concerned about the adding of the strong name to the mysql package, that is a breaking change. I might end up bumping to v4 so we can do this. A review from you and any thoughts welcome :)

PR #81, opened 2015-02-04
Support for selection of sql file encoding

dia0369:
In our project, we're using sql files with unicode characters. These need to be run against a MySQL database. Turns out, using Encoding.Default when defining the StreamReader (as DbUp does) doesn't correctly do that and the data inserted in the tables is gibberish. Managed to fix the issue by specifying Encoding.UTF8 when defining the StreamReader, but there was no built-in way of doing that with DbUp.
This change aims to add support for specifying custom encoding.

### DIFF
--- a/src/DbUp/Builder/StandardExtensions.cs
+++ b/src/DbUp/Builder/StandardExtensions.cs
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Reflection;
+using System.Text;
+using DbUp;
 using DbUp.Builder;
 using DbUp.Engine;
 using DbUp.Engine.Output;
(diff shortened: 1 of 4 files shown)


JakeGinnivan:
Nice addition

brcarey:
@JakeGinnivan any idea when this might make into a release?

JakeGinnivan:
I should be able to get it out this weekend.

JakeGinnivan:
I just am not sure what else will be in the release, need to just put together release notes and make sure there is nothing breaking etc.

Also want to look at the outstanding PR's. Aiming to do all that on saturday

brcarey:
Thanks mate, much appreciated

PR #85, opened 2015-04-01
Making SqlScriptExecutor inherit friendly

bradyholt:
By removing sealed from class definition and adding virutual on Execute, Log
methods, this class can now be inherited from and additional features can be built
on top of SqlScriptExecutor.

In particular, a useful subclass would be one that provides object definition scripting after scripts are executed (so you can have nice definition diffs) but I can imagine other useful scenarios as well.


JakeGinnivan:
Can't see any reason why not if it is useful to someone

PR #91, opened 2015-04-29
Fix sqllite

### LINKED ISSUE #90: DbUp.SqlLite.dll missing
Pretty sure that installing the DbUp.SqlLite Nuget package shooudl include this dll - as without it, can;t really do anything :)

dazinator:
I think this should fix #90 

I didn't do a fast forward or whatever it's called so if there is a "merge" commit in this that annoys you, feel free just to update the build file directly and i'll scrap this PR.


JakeGinnivan:
Am switching build to AppVeyor and redoing the packaging bit. Will get a new release out hopefully this weekend.

PR #93, opened 2015-04-30
Sq lite mono

dazinator:
I tried to use the existing sqllite provider within a xamarin android project (mono).

It failed at runtime.

I created a new / copy of the existing sql lite provider, but compiled against `mono.data.sqllite`. There are only a few minor code differences when compiling against mono.data.sqllite instead of system.data.sqllite but nothing major - just minor class / method name differences.

I have adjusted the build script accordingly to produce a `DbUp.SQLite.Mono` nuget package - but havent been able to test the build script - perhaps you team city will pick this PR up and do that?


JakeGinnivan:
@dazinator could we just use the Mono driver for everyone?

JakeGinnivan:
Rather than two packages. Otherwise we should have framework dependencies so we have a single nuget which depending on the platform you install for it has a different dependency. Then use compiler directives to change the namespace/types and we have two builds.

dazinator:
Hi Jake - I think the two build approach with compiler directives would
make sense. You may have to have conditional references if you wanted to
keep the single project file though so that when building the project for
mono the project file references the mono dependencies rather than full dot
net. Conditional references in project file (the condition being current
build configuration) may not play nicely with NuGet package management
features so may be better to have 2 seperate project files where the mono
project just has the code files added as linked files perhaps? and then use
compiler directives. As an excercise I have managed to compile dbup for
mono by installing the mono equivalent dependencies and then altering the
code in several places (which would need to become compiler directives when
done properly) and I recorded the problem areas I needed to alter in an
issue with a plan to sort it properly at a later date :)

dazinator:
Thinking further on this, I think making DbUp a PCL project might be the
better way to go. Will investigate further.

PR #95, opened 2015-05-06
Enable loading scripts from multiple assemblies

TrevorPilley:
I've been playing with DbUp in a project but in my scenario I need to manage scripts within multiple assemblies.

I have a service which assembly loads plugins and each of those plugins need their own table schema(s) maintaining.

This change allows me to manage the scripts for each plugin within the plugin project itself rather than the hosting assembly needing to know about any of the database schema for a plugin,

### DIFF
--- a/src/DbUp/Builder/StandardExtensions.cs
+++ b/src/DbUp/Builder/StandardExtensions.cs
@@ -425,4 +425,59 @@ public static class StandardExtensions
         return builder;
     }
 
+    /// <summary>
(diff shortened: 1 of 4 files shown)


TrevorPilley:
I'm not sure why it's showing as many changes in some of the files as it is - maybe a line ending issue or something like that?

JakeGinnivan:
Will merge changes via #112, thanks for the PR

PR #97, opened 2015-06-09
Issue-96-When-error-in-script-occurred-DbUp-method-Perform-Upgrade-do…

MaksymDrach:
When i invoke method PerformUpgrade (DbUp.Engine.UpgradeEngine), and i have script with error, DbUp fails and method PerformUpgrade returned DatabaseUpgradeResult which contain Exception, but info about exception does not provide the name of the failed script.

### DIFF
--- a/src/DbUp/Engine/UpgradeEngine.cs
+++ b/src/DbUp/Engine/UpgradeEngine.cs
@@ -62,6 +62,9 @@ namespace DbUp.Engine
         public DatabaseUpgradeResult PerformUpgrade()
         {
             var executed = new List<SqlScript>();
+
+            string executedScriptName = null;
+
             try
             {
                 using (configuration.ConnectionManager.OperationStarting(configuration.Log, executed))
@@ -80,6 +83,8 @@ namespace DbUp.Engine
 
                     foreach (var script in scriptsToExecute)
                     {
+                        executedScriptName = script.Name;
+
                         configuration.ScriptExecutor.Execute(script, configuration.Variables);
 
                         configuration.Journal.StoreExecutedScript(script);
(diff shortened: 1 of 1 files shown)


src/DbUp/Engine/UpgradeEngine.cs
```diff
@@ -93,6 +98,7 @@ public DatabaseUpgradeResult PerformUpgrade()
             }
             catch (Exception ex)
             {
+                ex.Data.Add("Error occurred in script: ", executedScriptName);
                 configuration.Log.WriteError("Upgrade failed due to an unexpected exception:\r\n{0}", ex.ToString());
```
JakeGinnivan: Does the data get written out when you ToString the exception?

MaksymDrach:
For sure no, but ex.ToString() used for add info to log, in this case by default for logging used Trace, and there already exist message about crashed script( Execute method did it internally).
But I need that info  about crashed script will be in the exception object, because the exception object(ex) returned from the catch construction:
return new DatabaseUpgradeResult(executed, false, ex);

MaksymDrach:
Sorry, accidentally closed.

src/DbUp/Engine/UpgradeEngine.cs
```diff
@@ -93,6 +98,7 @@ public DatabaseUpgradeResult PerformUpgrade()
...
                 return new DatabaseUpgradeResult(executed, false, ex);
```
MaksymDrach: Here the ex parameter contain all info about exception and now provide also info about the crashed script in the ex object.

src/DbUp/Engine/UpgradeEngine.cs
MaksymDrach: For sure no, but ex.ToString() used for add info to log, in this case by default for logging used Trace, and there already exist message about crashed script( Execute method did it internally).
But I need that info about crashed script will be in the exception object, because the exception object(ex) returned from the catch construction:
return new DatabaseUpgradeResult(executed, false, ex);

Issue #74, opened 2014-11-26
Create Db if it doesn't exist


llaslo:
What I'd really like the is the ability to have DbUp have an option to create the database if it doesn't exist.  I tried to solve this by having my first script do this:

**0001_InitialCreate.sql**

    IF NOT EXISTS ( SELECT [Name] FROM sys.databases WHERE [name] = 'MyDatabase' )
    BEGIN
	    CREATE DATABASE [MyDatabase]
    END
    GO

    USE [MyDatabase]
    GO

    IF NOT EXISTS ( SELECT * FROM sysobjects where Name='MyTable' and xtype='U')
	    CREATE TABLE dbo.MyTable (
		    [Id] INT NOT NULL PRIMARY KEY IDENTITY,
		    [Name] VARCHAR(50) NOT NULL,
		    [CreatedBy] VARCHAR(50) NOT NULL,
		    [CreatedDate] DATETIME2(7) NOT NULL DEFAULT SYSDATETIME())
    GO

Subsequent scripts look like this:

**0002_AddAgeColumnToMyTable.sql**

    USE [MyDatabase]
    GO

    ALTER TABLE dbo.MyTable
    ADD [Age] INT NULL
    GO
    
This approach works fine if you leave the database name out of the connection string, and just use *USE* statements in your scripts.  The one problem, though, is that it doesn't find the SchemaVersions table so it re-runs every script every time.  (although it does actually find the table and insert the scripts into it).

worldspawn:
```
using (var con = new SqlConnection(masterConnectionString))
            {
                using (var cmd = new SqlCommand())
                {
                    cmd.Connection = con;
                    con.Open();

                    cmd.CommandText = string.Format("SELECT COUNT([name]) FROM [sys].[databases] WHERE [name] = N'{0}'", database);
                    var databaseExists = Convert.ToBoolean(cmd.ExecuteScalar());
                    if (!databaseExists) CreateDatabase(con, database);
                    con.Close();
                }
}

public static void CreateDatabase(SqlConnection con, string database)
        {
            Console.Write("Creating database...");

            using (var cmd = new SqlCommand())
            {
                cmd.Connection = con;
                cmd.CommandText = string.Format("CREATE DATABASE {0}", database);
                cmd.ExecuteNonQuery();
            }

            Console.WriteLine("Done");
        }
```

JakeGinnivan:
[...]

PR #99, opened 2015-07-28
first stab at  way to have DbUp create a db if it doesn't exist.

aggieben:
This is a first attempt at #74.


aggieben:
@JakeGinnivan I went ahead and wrote this up, although TBH, I think I've found a better way for my particular needs.  I plan to do the db creation in the console program consuming DbUp; that gives me more control.  Feel free to either run with this or discard it as you see fit :smile:

JakeGinnivan:
I think I will close this, I think it is better to have this sort of code in the console program or something. Rather than needing DbUp to know about system connections and such. Thanks anyways!

PR #102, opened 2015-08-07
Fixes #101

### LINKED ISSUE #101: Variable substitution doesn't seem to be working
I'm using the latest release on nuget (3.2.1, right?).  I set up my builder like this:

```csharp
            var engine = DeployChanges.To
                .PostgresqlDatabase(_settings.ConnectionStrings[_settings.Connection])
                .WithScriptsEmbeddedInAssembly(Assembly.GetExecutingAssembly(), (path) => FilterScript(path))
                .LogScriptOutput()
                .LogToConsole()
                .WithTransactionPerScript()
                .WithCustomEnvironment()
                .WithVariablesEnabled()
                .Build();
```

`WithCustomEnvironment()` is an extension method in my solution (names changed here manually, please excuse a typo if I made one):

```csharp
namespace Migrations
{
    using System;
    using System.Collections.Generic;
    using DbUp.Builder;
    public static class UpgradeEngineBuilderExtensions
    {
        public static UpgradeEngineBuilder WithCustomEnvironment(this UpgradeEngineBuilder builder)
        {
            var envChannelName = Environment.GetEnvironmentVariable("CHANNEL_NAME");
            if (string.IsNullOrWhiteSpace(envChannelName))
                envChannelName = "sbcomall";

            var envBlah = Environment.GetEnvironmentVariable("BLAH");
            if (string.IsNullOrWhiteSpace(envBlah ))
                envBlah = "0000";

            var envDeeDah = Environment.GetEnvironmentVariable("DEEDAH");
            if (string.IsNullOrWhiteSpace(envDeeDah))
                envDeeDah= "aaa";


            builder.WithVariables(new Dictionary<string, string> {
                {"CHANNEL_NAME", envChannelName},
                {"BLAH", envBlah},
                {"DEEDAH", envDeeDah},
            });

            return builder;
        }
    }
}
```

In my SQL scripts, I'm using the variables like this:

```sql
-- Function notify_insert()
--   Can be invoked in an insert trigger for any table
--   will notify subscribers of insert event and send JSON of inserted row
CREATE OR REPLACE FUNCTION notify_insert() RETURNS TRIGGER AS $$
DECLARE
BEGIN
    PERFORM (
        pg_notify('$CHANNEL_NAME$',
                  CAST((SELECT row_to_json(row)
                          FROM (SELECT 'insert' AS Command,
                                            $BLAH$ AS Blah,
                                            $DEEDAH$ AS DeeDah,
                                            NEW) row)
                        AS VARCHAR)
        )
    );
    RETURN NEW;
END;
$$ LANGUAGE plpgsql;
```

When I execute the engine, the Npgsql provider complains of an unterminated dollar-quoted string, which is particular to Postgresql syntax.  However, it means that the variables weren't replaced (and the exception shows the unreplaced variables on the console).

I tried explicitly running the `VariableSubstitutionPreprocessor` on the scripts, and the substitutions worked fine (I just printed the contents to the console after running `.Process()` on each script's content, so I'm not sure I understand why it wouldn't work when a migration is actually running.

Ideas?

aggieben:
Enable variable substitution in Postgresql migrations ;-)

### DIFF
--- a/src/DbUp.Postgresql/PostgresqlExtensions.cs
+++ b/src/DbUp.Postgresql/PostgresqlExtensions.cs
@@ -36,7 +36,7 @@ public static class PostgresqlExtensions
     {
         var builder = new UpgradeEngineBuilder();
         builder.Configure(c => c.ConnectionManager = connectionManager);
-        builder.Configure(c => c.ScriptExecutor = new SqlScriptExecutor(() => c.ConnectionManager, () => c.Log, null, () => false, c.ScriptPreprocessors));
+        builder.Configure(c => c.ScriptExecutor = new SqlScriptExecutor(() => c.ConnectionManager, () => c.Log, null, () => c.VariablesEnabled, c.ScriptPreprocessors));
         builder.Configure(c => c.Journal = new PostgresqlTableJournal(() => c.ConnectionManager, () => c.Log, null, "schemaversions"));
         builder.WithPreprocessor(new PostgresqlPreprocessor());
         return builder;


dazinator:
Nice one - hopefully this can be merged soon!

If you get a chance, and you are feeling particularly generous, it might also be worth creating a regression test for this one! There is already a very similar test for variable substitution but for the SqlServer provider that you could potentially use a a basis for the Postgres test: https://github.com/DbUp/DbUp/blob/master/src/DbUp.Tests/DeployChangesBuilderTests.cs#L13

The SqlServer based test is able to mock up the underlying `IDbConnection` but the Postgres extensions / provider API doesn't allow for this, so you would might have to change a few things there to improve its testability if you did decide to have a go at this :)

aggieben:
I'll see if I can find time for it.  If you'll assign me this issue then I'll have it on my issue list and will be less likely to forget.

JakeGinnivan:
FYI I am working on making testing easier, so once I have the infrastructure in place I will get this in and have a test around it

JakeGinnivan:
Ok got a test running which proves variable substitution is broken for Firebird, MySql and Postgresql. Thanks for the PR

PR #113, opened 2015-09-06
My sql refactor

JakeGinnivan:
#75 rebased and a few extra changes to use the new way of testing


namzat:
Hey Jake, you mentioned concern over adding the strong name to the MySQL package.  It has been a while, what exactly are you referring to?  I went back the parent of where I cloned (commit d3f4363) and looked and all of the projects were strong named.  I'm not sure if you mean the MySQL project I added or Nuget package or what exactly you mean by package  Can you shed some light on this for me?  Thanks! :)

JakeGinnivan:
@namzat https://github.com/DbUp/DbUp/pull/113/files#diff-798c325b17e29c5609271bb0ebc1b281R41

Other projects are strong named, MySql was not. Strong naming is a breaking change because you cannot have any transitive references which are not recompiled when you add the strong name.

Honestly I think Strong Naming should be dropped entirely from DbUp

JakeGinnivan:
@namzat I have removed strong naming for now from the MySql project to not block a release with a breaking change

namzat:
Fine with me.  I'm not sure why I strong named the MySql project anyway except perhaps that the other projects were strong named.  I would agree with removing strong naming from DbUp entirely.  Not sure that it provides much benefit.

PR #114, opened 2015-09-07
Added EnsureDatabase.For

rickculpepper:
# EnsureDatabase.For #

Issue **#74 Create Db if it doesn't exist**

This new DbUp method ensures the existence of the database that is going to be upgraded. If the database does not exist, it is created. The `CREATE DATABASE` command is executed with default values for all options.

## Changes ##

1. Added a new static class `EnsureDatabase` with a property, `EnsureDatabase.For`, for use semantics similar to `DeployChanges.To`. However, the type of `EnsureDatabase.For` is a new class, `SupportedDatabasesForEnsureDatabase` to disambiguate the extension classes. 
2. Added extension methods for all supported databases, but only implemented the SQL Server extension method. All other extensions methods throw `System.NotImplementedException`.

## Notes ##
1. The argument connectionString is the same connection string used with the Builder for the target database. The extension method works out the connection string for `master`.
2. The return type of the extension method is void. If the database engine throws an exception when creating the database, that exception is _**not**_ caught.

## Example ##
```
EnsureDatabase.For.SqlDatabase(connectionString);
```
## Open Issues ##
It seems like `EnsureDatabase.For` should let the world know what it is doing. For this reason, there are two TODO: items in the SQL Server extension method that highlight code that may need to be changed. The challenge is that we are not in the Builder, so we have no logging context.

* `SqlServerExtensions.cs`, line 129-130. There is a TODO and a `Debug.WriteLine` that displays the master database connection string. This could probably be removed, but seemed like a good idea during debugging.
* `SqlServerExtensions.cs`, line 171-172. There is a TODO and a `ConsoleWriteLine` where the extension method announces that it created the database. This seems like a good idea. However,  `Console.Writeline()` seems too invariant, but the lack of a logging context doesn't leave to many (easy) options.

### DIFF
--- a/src/DbUp.Firebird/FirebirdExtensions.cs
+++ b/src/DbUp.Firebird/FirebirdExtensions.cs
@@ -42,15 +42,4 @@ public static class FirebirdExtensions
         builder.WithPreprocessor(new FirebirdPreprocessor());
         return builder;
     }
-
-    /// <summary>
-    /// Ensures that the database specified in the connection string exists.
-    /// </summary>
-    /// <param name="supported">Fluent helper type.</param>
-    /// <param name="connectionString">The connection string.</param>
-    /// <returns></returns>
-    public static void FirebirdDatabase(this SupportedDatabasesForEnsureDatabase supported, string connectionString)
-    {
-        throw new NotImplementedException("EnsureDatabase not supported for Firebird databases.");
-    }
 }
--- a/src/DbUp.Firebird/FirebirdExtensions.cs
+++ b/src/DbUp.Firebird/FirebirdExtensions.cs
@@ -1,5 +1,6 @@
 ﻿using DbUp.Builder;
 using System;
+using DbUp;
 using DbUp.Firebird;
 using DbUp.Engine.Transactions;
 using DbUp.Support.SqlServer;
@@ -40,5 +41,16 @@ public static class FirebirdExtensions
         builder.Configure(c => c.Journal = new FirebirdTableJournal(() => c.ConnectionManager, () => c.Log, "schemaversions"));
         builder.WithPreprocessor(new FirebirdPreprocessor());
         return builder;
+    }
+
+    /// <summary>
+    /// Ensures that the database specified in the connection string exists.
+    /// </summary>
+    /// <param name="supported">Fluent helper type.</param>
+    /// <param name="connectionString">The connection string.</param>
+    /// <returns></returns>
+    public static void FirebirdDatabase(this SupportedDatabasesForEnsureDatabase supported, string connectionString)
+    {
+        throw new NotImplementedException("EnsureDatabase not supported for Firebird databases.");
     }
 }
--- a/src/DbUp/Support/SqlServer/SqlServerExtensions.cs
+++ b/src/DbUp/Support/SqlServer/SqlServerExtensions.cs
@@ -132,20 +132,25 @@ public static class SqlServerExtensions
 
         if (logger == null) throw new ArgumentNullException("logger");
 
-        var connectionStringBuilder = new SqlConnectionStringBuilder(connectionString);
+        var masterConnectionStringBuilder = new SqlConnectionStringBuilder(connectionString);
 
-        var databaseName = connectionStringBuilder.InitialCatalog;
+        var databaseName = masterConnectionStringBuilder.InitialCatalog;
 
         if (string.IsNullOrEmpty(databaseName) || databaseName.Trim() == string.Empty)
         {
             throw new InvalidOperationException("The connection string does not specify a database name.");
         }
 
-        connectionStringBuilder.InitialCatalog = "master";
+        masterConnectionStringBuilder.InitialCatalog = "master";
 
-        logger.WriteInformation("Master ConnectionString => {0}", connectionStringBuilder.ToString());
+        var logMasterConnectionStringBuilder = new SqlConnectionStringBuilder(masterConnectionStringBuilder.ConnectionString)
+        {
+            Password = String.Empty.PadRight(masterConnectionStringBuilder.Password.Length,'*')
+        };
+        
+        logger.WriteInformation("Master ConnectionString => {0}", logMasterConnectionStringBuilder.ConnectionString);
 
-        using (var connection = new SqlConnection(connectionStringBuilder.ConnectionString))
+        using (var connection = new SqlConnection(masterConnectionStringBuilder.ConnectionString))
         {
             connection.Open();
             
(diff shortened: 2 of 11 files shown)


rickculpepper:
BTW, the DbUp.Tests project would not compile when I got it. The compile errors had to do with automatic properties that did not have a setter... ???

src/DbUp.Firebird/FirebirdExtensions.cs
```diff
@@ -41,4 +42,15 @@ public static UpgradeEngineBuilder FirebirdDatabase(IConnectionManager connectio
...
         return builder;
     }
+
+    /// <summary>
+    /// Ensures that the database specified in the connection string exists.
+    /// </summary>
+    /// <param name="supported">Fluent helper type.</param>
+    /// <param name="connectionString">The connection string.</param>
+    /// <returns></returns>
+    public static void FirebirdDatabase(this SupportedDatabasesForEnsureDatabase supported, string connectionString)
```
JakeGinnivan: Remove all the extension methods which are not implemented. As people submit pull requests they can be added

src/DbUp/Support/SqlServer/SqlServerExtensions.cs
```diff
@@ -97,4 +101,77 @@ public static UpgradeEngineBuilder JournalToSqlTable(this UpgradeEngineBuilder b
...
         return builder;
     }
+
+    /// <summary>
+    /// Ensures that the database specified in the connection string exists.
+    /// </summary>
+    /// <param name="supported">Fluent helper type.</param>
+    /// <param name="connectionString">The connection string.</param>
+    /// <returns></returns>
+    public static void SqlDatabase(this SupportedDatabasesForEnsureDatabase supported, string connectionString)
```
JakeGinnivan: I would split this into `server` and `database`, it is a much clearer API

JakeGinnivan: Actually ignore that. This is fine

JakeGinnivan: Add `, IUpgradeLog logger` to this method.

You can provide an overload which will log to console by default, but it should be overridable.

JakeGinnivan:
If you rebase onto master, I have removed the use of c#6 features which require VS2015 to compile.

src/DbUp.Firebird/FirebirdExtensions.cs
rickculpepper: Will do.

src/DbUp/Support/SqlServer/SqlServerExtensions.cs
rickculpepper: Glad you changed your mind on replacing `connectionString`. That seemed to me that it fits the DbUp use cases... much better than `server` and `database`.

Thanks for the tip on `IUpgradeLog logger`. Will do...

rickculpepper:
@JakeGinnivan not familiar with rebasing, so I read up a bit. I tried to get this done, but I'm not sure if I did it right or not. Also, ReSharper "helped" me turn on C#6 support, but now I don't see how to turn it off... :P

rickculpepper:
Never mind. I see the PR picked up my commit with these changes.

rickculpepper:
@JakeGinnivan, I can no longer see the appveyor output to see what is failing in the build. Did something change? Also, what is the meaning of the message "This branch has conflicts that must be resolved"? Is there something that I can do here?

JakeGinnivan:
Master has moved on, you need to rebase this branch.

```
git fetch upstream
git rebase upstream/master
git push origin CreateDatabase -f
```

You should get a build then

rickculpepper:
Thanks for the help, @JakeGinnivan! 

However, I think I might need a little more help here. :confused: Let me explain...

In my clone, the remote is named DbUp rather than upstream. Not quite sure how that happened, but the command `git remote –v` produces the following output:

```
DbUp    https://github.com/DbUp/DbUp.git (fetch)
DbUp    https://github.com/DbUp/DbUp.git (push)
origin  https://github.com/rickculpepper/DbUp.git (fetch)
origin  https://github.com/rickculpepper/DbUp.git (push)
```
So, I did the following.

```
git fetch DbUp
```
This appeared to succeed. Next I did this...
```
git rebase DbUp/master
```
This comand failed with a conflict:
```
CONFLICT (modify/delete): src/DbUp.SQLite.Mono/SqliteExtensions.cs deleted in HEAD and modified in {{ChangesetDescription}}
Failed to merge in the changes.
Patch failed at 0001 {{ChangesetDescription}}
The copy of the patch that failed is found in:
   c:/github/rickculpepper/DbUp/.git/rebase-apply/patch

When you have resolved this problem, run "git rebase --continue".
If you prefer to skip this patch, run "git rebase --skip" instead.
To check out the original branch and stop rebasing, run "git rebase --abort".
```
I looked at the patch file... and it's too much for this newbie.

So, I ran `git rebase --abort` to get back to square one.

Gonna try a couple of things here to see if I can get this working, but I may need a little mentoring (intervention?)... Any advice anyone can offer would be greatly appreciated.

Thanks!

JakeGinnivan:
You run `git mergetool` to resolve conflicts

JakeGinnivan:
I have created a gitter room as well, jump on there and I can help - https://gitter.im/DbUp/DbUp

src/DbUp/Support/SqlServer/SqlServerExtensions.cs
```diff
@@ -97,4 +102,91 @@ public static UpgradeEngineBuilder JournalToSqlTable(this UpgradeEngineBuilder b
...
+        var databaseName = connectionStringBuilder.InitialCatalog;
+
+        if (string.IsNullOrEmpty(databaseName) || databaseName.Trim() == string.Empty)
+        {
+            throw new InvalidOperationException("The connection string does not specify a database name.");
+        }
+
+        connectionStringBuilder.InitialCatalog = "master";
+
+        logger.WriteInformation("Master ConnectionString => {0}", connectionStringBuilder.ToString());
```
JakeGinnivan: Does this show the password?

rickculpepper: :cry: 
Yes, it does. Good catch. I'll fix it shortly.

src/DbUp/Support/SqlServer/SqlServerExtensions.cs
rickculpepper: Fixed...

JakeGinnivan:
Thanks! This should be merged by the time you get up :)

PR #117, opened 2015-09-10
Sqlite fixes

JakeGinnivan:


JakeGinnivan:
Fyi @dazinator 

The mono package only targets .net 4.0 - we probably need to expand it to build against multiple targets so it actually works with xamarin and such?

dazinator:
Yeah, we need builds for targets: MonoTouch, MonoDroid, and Mono.
That will involve adding some conditional compilation statements to the DbUp (core) library.  
I have done this locally for an android project, in a hacky way, but it needs all doing properly!
Will create a branch and do some work around this at the weekend hopefully :)

dazinator:
I am guessing though that this may then break the builds on the build server if those frameworks aren't installed?

dazinator:
- I think this may mean you probably need to install xamarin on the build server... Not sure how you feel about that.. Pretty sure xamarin offer open source licences: https://resources.xamarin.com/open-source-contributor.html

PR #125, opened 2015-09-12
Sqlcmd script processing

rickculpepper:
Pull Request for issue #123

Issue #124, opened 2015-09-11
FileSystemScriptProvider Support for Folder Recursion


hrc3:
## Story

As a DbUp User, I would like to be able to use upgrade scripts from the file system in a multi-level folder structure in order to facilitate better organization of the upgrade scripts.
## Explanation

When upgrade scripts are included in the DbUp project as internal resources, the Scripts folder can contain multiple levels of sub-folders to help organize the upgrade scripts. However, when upgrade scripts are provided from a file system folder that does not contain scripts marked as resources, all the scripts must be in the root of the specified folder. Any sub-folders will be ignored.

The path to the file relative to the specified starting point should be used as the "manifest name" for the script. This means that sub-folder names will be significant in the ordering of scripts for processing.

rickculpepper:
@JakeGinnivan please mark as an enhancement for 4.0 and assign to me

rintje:
Would love to see this enhancement too. 
Its just what we are missing to implement dbup with our db script folder layout.

cmarrades:
I just found out if I split the scripts in multiple folders, it won't execute them ordered by script name. Not sure how it works, but I think this story would help to solve my problem as well.

droyad:
This is now supported

cmarrades:
better late than never :P 

Thx

InvincibleDRT:
with the new version i am getting the following error "SqlConnection does not support parallel transactions"
My code works fine with 3.3.5 build

droyad:
@InvincibleDRT Could you share your DbUp builder code as well a the stack trace for that error?

InvincibleDRT:
string branchDBInfo = string.Format(ConfigurationManager.AppSettings[keyConn], branchInfo);
            EnsureDatabase.For.SqlDatabase(branchDBInfo);

            FileSystemScriptOptions fsso = new FileSystemScriptOptions();
            fsso.IncludeSubDirectories = true;
            var upgradeBuilder = DeployChanges.To
                                    .SqlDatabase(branchDBInfo)
                                    .WithScriptsFromFileSystem(path, fsso)
                                    .WithVariablesDisabled().LogTo(this);


            upgradeBuilder.Configure(x => x.ScriptExecutor.ExecutionTimeoutSeconds = 60 * 60);
            var upgrader = upgradeBuilder.WithTransactionPerScript().Build();

            var result = upgrader.PerformUpgrade();

            object message = (result.Successful) ? "Success" : (object)result.Error;
[...]

PR #128, opened 2015-09-12
Script folder recursion

rickculpepper:
Pull Request for issue #124.

This Pull Request does _not_ contain any breaking changes to the public API. However, it _**does**_ contain a _**BREAKING CHANGE**_ for the script names stored in the DbUp journal table.

In order to support folder recursion, the `SqlScript.Name` property must contain the path to the script, not just the file name. This changes the script name that is stored in the journal table. So, using this version of DbUp to update an existing database where scripts are source from the file system will make a mess unless the file-based script names in the journal table are first updated to reflect the new script name which includes the relative path from the root script folder.

It seems like this ought to be mentioned in `journaling.md`, but I wasn't quite sure how to introduce it.

### DIFF
--- a/src/DbUp.Tests/DbUp.approved.cs
+++ b/src/DbUp.Tests/DbUp.approved.cs
@@ -263,9 +263,13 @@ namespace DbUp.ScriptProviders
     public class FileSystemScriptProvider : DbUp.Engine.IScriptProvider
     {
         public FileSystemScriptProvider(string directoryPath) { }
+        public FileSystemScriptProvider(string directoryPath, bool recursive) { }
         public FileSystemScriptProvider(string directoryPath, System.Func<string, bool> filter) { }
+        public FileSystemScriptProvider(string directoryPath, System.Func<string, bool> filter, bool recursive) { }
         public FileSystemScriptProvider(string directoryPath, System.Text.Encoding encoding) { }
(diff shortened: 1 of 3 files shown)


JakeGinnivan:
We need to fix your master branch up so it doesn't have all that history. From memory your remote was `DbUp`? 

First lets fix this pull request:

```
git fetch DbUp
git rebase DbUp/master -i
```

This will do an interactive rebase, you want to delete all lines except for your actual commit out of that file which pops up. Save the file and close the editor. It will then rebase your actual commit.

If you `git push origin ScriptFolderRecursion -f` you will see you will go from 11 commits down to 1.

Now to fix master up:

```
git checkout master
git reset DbUp/master --hard
```

Now you have a clean master. To update master with changes from upstream

```
git checkout master
git fetch DbUp
git merge DbUp/master --ff-only
```

`--ff-only` prevents git from creating a merge commit, so if you accidently ever commit to master then that command will fail. In that case you need to branch master to a feature branch, rebase that feature branch then reset master back to what the upstream master is.

Hope that all makes sense :)

JakeGinnivan:
If you could fix #125 too, that would be great. If you get stuck just ping me on gitter

JakeGinnivan:
I can also fix on my side when merging in, but that will not sort your master branch out

rickculpepper:
@JakeGinnivan I am closing this because I realized I missed a couple of things and I want to add a sample application.

Back soon!

Issue #123, opened 2015-09-11
SQLCMD Script Preprocessor


rickculpepper:
As a DBA who uses SQLCMD variables in my SQL scripts, I would like to have a DbUp Preprocessor to replace SQLCMD variables in generated scripts so that I do not have to manually replace SQLCMD variable references with DbUp variable references.

For example, DbUp expects a variable reference in the form `$variableName$`. SQLCMD variable references are in the form `$(variableName)`.

This is _**not**_ a request to support full-on SQLCMD preprocessing for commands such as the following.
```
:setvar VariableName "VariableValue"
:r c:\scripts\myscriptfile.sql
```
In fact, _**none**_ of the SQLCMD commands should be supported or processed in any way. Any line in a SQL script which begins with a colon character should be converted to a comment by adding "-- " to the beginning of the line.

rickculpepper:
My apologies for three pull requests on this issue, @JakeGinnivan... that's my inexperience again. I realize now that, instead of editing directly on GitHub, I should have done the editing locally in the same branch as the code changes... then you'd just have one pull request to deal with. :stuck_out_tongue_winking_eye: Next time for sure.

The good news is that I think I got the code changes right this time. All the unit tests are green and I even added a new BDDfy test for the sqlcmd variable substitution.

Looking forward to your feedback.

rickculpepper:
@JakeGinnivan please assign to me

jamiebarrow:
sqlcmd support would be great. bumping this :)

HasAndries:
It would be awesome if SQLCMD was supported. Is there any chance of reviving this?

michalporeba:
Well spotted @droyad, I'll have a look what was done previously, it seems like some work has been done, but it was too much for a single PR. I'll see if I can make something for the SQLCMD only.

dazinator:
Any more news on this? I have a question about how this is meant to work:

> sqlcmd supports a number of commands. These commands are processed by sqlcmd before the Transact-SQL script is submitted to the SQL Server engine for processing.

> DbUp should not attempt to support the functionality of the sqlcmd commands. Rather, DbUp should convert every sqlcmd command to a double-dash comment and return it as part of the content of the script to be processed by SQL Server.

So... given that DbUp will actuall comment out the `sqlcmd` commands before submitting them via ADO.Net, - where is the bit that  actually sends the `sqlcmd` commands to `sqlcmd` - will DbUp do this also? How does that bit work?

[...]

PR #129, opened 2015-09-14
SQLCMD Support and File Folder Recursion for Scripts

ternaryalchemy:
This pull request addresses both of the following Issues. The pull includes new unit tests to verifyy and a new sample application to demonstrate the functionality.

#123 SQLCMD Script Support
#124 FileSystemScriptProvider Support for Folder Recursion

- [X] Update `src/DbUp.Tests/ScriptProvider/FileSystemScriptProviderTests.cs` method `the_file_should_contain_content()` to use `.ShouldBeTrue(...)`.
- [X] Remove Solution Any CPU configuration.
- [X] Remove commented line from `src/DbUp/Engine/Preprocessors/StripSchemaPreprocessor.cs` method `Process(string contents)`.
- [X] Revise formatting of `src/DbUp/ScriptProviders/FileSystemScriptProvider.cs` method 'GetScripts(..)` considering comments made by @JakeGinnivan
- [ ] Move changes to `src/DbUp/Support/SqlServer/SqlCommandReader.cs` to new class, `src/DbUp/Support/SqlServer/MSSqlCommandReader.cs'. Model the new class after ?????
- [ ] Refactor File Folder Recursion code to avoid breaking change.
- [ ] Refactor sample app `SqlServer.FolderRecursion`.
  - [ ] Create new sample app for Folder Recursion that is database-independent, i.e., separate from SQLCMD features
  - [ ] `SqlServer.FolderRecursion` should 
- [ ] Update docs to ref
- [ ] Revise Regex pattern in `StripSchemaPreprocessor.cs`. Current pattern matches `$(schema$'. If needed, revert to separate RegEx patterns for DbUp- and SQLCMD-style variable recognition.
- [ ] Consider changing `VariableSubstitutionPreprocessor.cs` to use a single Regex pattern to identify both DbUp- and SQLCMD-style variables -- without introducing the problem mentioned above.

@JakeGinnivan, see my email for more thoughts on this PR. Let me know if you want me to add that commentary here.


src/DbUp.Tests/ScriptProvider/FileSystemScriptProviderTests.cs
```diff
@@ -99,6 +98,138 @@ public void encoding_reader_is_correct()
...
+            {
+                Assert.AreEqual(expectedFileCount, filesToExecute.Count());
+            }
+
+            [Then]
+            public void the_file_should_contain_content()
+            {
+                foreach (var sqlScript in filesToExecute)
+                {
+                    Assert.IsTrue(sqlScript.Contents.Length > 0);
```
JakeGinnivan: `(sqlScript.Contents.Length > 0).ShouldBeTrue(sqlScript)` will be much more informative whe/ifn this fails

src/DbUp.sln
```diff
@@ -37,60 +37,118 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DbUp.MySql", "DbUp.MySql\Db
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DbUp.SQLite.Mono", "DbUp.SQLite.Mono\DbUp.SQLite.Mono.csproj", "{5C474B5C-019C-4C03-9E20-2D0CDEAAEA7B}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "SqlServer.FolderRecursion", "Samples\SqlServer.FolderRecursion\SqlServer.FolderRecursion.csproj", "{BFF56C02-6754-4086-8319-461B47E5FFC3}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
```
JakeGinnivan: Can we remove the Any CPU configuration again?

src/DbUp/Engine/Preprocessors/StripSchemaPreprocessor.cs
```diff
@@ -8,12 +9,18 @@ namespace DbUp.Engine.Preprocessors
...
+        private static readonly Regex dbUpStyleSchemaRegex = new Regex(@"\$schema\$\.", RegexOptions.IgnoreCase);
+        private static readonly Regex sqlcmdStyleSchemaRegex = new Regex(@"\$\(schema\)\.", RegexOptions.IgnoreCase);
+
         /// <summary>
         /// Performs some proprocessing step on a script
         /// </summary>
         public string Process(string contents)
         {
-            return Regex.Replace(contents, @"\$schema\$\.", string.Empty, RegexOptions.IgnoreCase);
+            //return Regex.Replace(contents, @"\$schema\$\.", string.Empty, RegexOptions.IgnoreCase);
```
JakeGinnivan: Remove commented line

src/DbUp/Engine/Preprocessors/StripSchemaPreprocessor.cs
```diff
@@ -8,12 +9,18 @@ namespace DbUp.Engine.Preprocessors
     /// </summary>
     public class StripSchemaPreprocessor : IScriptPreprocessor
     {
+
+        private static readonly Regex dbUpStyleSchemaRegex = new Regex(@"\$schema\$\.", RegexOptions.IgnoreCase);
```
JakeGinnivan: Change to a single regex `\$(\(schema\)|schema\$)\.` maybe

src/DbUp/ScriptProviders/FileSystemScriptProvider.cs
```diff
@@ -56,24 +131,65 @@ public FileSystemScriptProvider(string directoryPath, Encoding encoding)
...
+            //  Notes:  
+            //      1.  When Directory.Getfiles() is passed an extension that is EXACTLY 3 characters in length, the 
+            //          resulting file list will contain any file with an extension longer than 4 characters that
+            //          begins with the specified extension. This is a documented FEATURE. 
+            //      2.  In order to limit the returned file list to just those with the extension ".sql" (nothing 
+            //          more and nothing less, it is necessary to add a .Where condition that filters on the 
+            //          extension.
+            //      Why do both? To limit the list returned by Directory.Getfiles() in case there are lots of other
+            //      files in the specified DirectoryPath...
+            return Directory
```
JakeGinnivan: Looks like a good change, can I suggest a formatting change:

``` csharp
var searchOptions = Recursive ? SearchOption.AllDirectories : SearchOption.TopDirectoryOnly;
return Directory
    .GetFiles(DirectoryPath, "*.sql", /* Note(1) above */, searchOptions)
    .Where(FileExtensionIsDotSql) //Note(2) above...
    .Where(Filter)
    .Select(ReadFile)
    .ToList();

SqlScript ReadFile(string filePath)
{
    // Not sure naming is correct, hard to understand exactly what is trying to be achieved here
    var rootFolder = Path.GetDirectoryName(DirectoryPath).Split('\\').LastOrDefault();
    var pathRelativeToSolutionDir = string.Concat(rootFolder, "\\", filepath.Replace(DirectoryPath, string.Empty));
    return SqlScript.FromStream(pathRelativeToSolutionDir, new FileStream(filepath, FileMode.Open, FileAccess.Read), Encoding);
}
bool FileExtensionIsDotSql(string filepath)
{
   return Path.HasExtension(filepath) && 
             ".sql".Equals(Path.GetExtension(filepath), StringComparison.InvariantCultureIgnoreCase);
}
```

src/DbUp/Support/SqlServer/SqlCommandReader.cs
```diff
@@ -88,6 +91,11 @@ private string ReadCommand()
                     ReadCustomStatement();
                     continue;
                 }
+                if (IsSqlCmdCommand)
```
JakeGinnivan: Be careful with this, I think this reader is used by other databases other than sql?

src/Samples/SqlServer.FolderRecursion/Program.cs
```diff
@@ -0,0 +1,81 @@
+﻿using System;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using DbUp;
+using DbUp.Helpers;
+
+namespace SqlServer.FolderRecursion
+{
+    class Program
```
JakeGinnivan: Is it possible to collapse into a single sample? Or even just move the info over to the docs only. Not convinced of the value of another sample app?

src/DbUp.Tests/ScriptProvider/FileSystemScriptProviderTests.cs
rickculpepper: Got it. Change made. FWIW, this is not code that I added, but happy to make the change.

src/DbUp.sln
rickculpepper: Now sure how this crept back in, but the Solution Configuration Manager seems to have a mind of it's own. This might have happened when I did a Batch Build is VS 2013. Anyway, it's gone now. :stuck_out_tongue_closed_eyes: ...

src/DbUp/Engine/Preprocessors/StripSchemaPreprocessor.cs
rickculpepper: Oops. Sorry. :S Done...

src/DbUp.sln
JakeGinnivan: When you added the sample project Visual Studio would have put it back in. Rather annoying

Sent from my Windows Phone
________________________________

src/DbUp/Engine/Preprocessors/StripSchemaPreprocessor.cs
rickculpepper: In the end, I'm going to use `\$\(?schema[\$\)]\.` It's a little more concise.

Truthfully, I went back and forth on this. I opted for the more verbose option because it seemed to indicate more clearly what was going on, i.e., "self-documenting code". However, I'm good with making it a single Regex pattern. Not a big deal. ;)

src/DbUp.sln
rickculpepper: Annoying indeed!

src/DbUp/Engine/Preprocessors/StripSchemaPreprocessor.cs
JakeGinnivan: Up to you, also feel free to leave it. Just a suggestion :)

src/DbUp/ScriptProviders/FileSystemScriptProvider.cs
rickculpepper: I like this refactoring, Jake. It makes the `return Directory...` much, much clearer.

As for the "What is trying to be achieved here..."

I am setting the Name property of the SqlScript object to the *relative* directory path. So, if DirectoryPath (set in the .ctor) is `C:\github\rickculpepper\DbUp\Samples\SqlServer.FolderRecursion\Scripts\` and the file is `C:\github\rickculpepper\DbUp\Samples\SqlServer.FolderRecursion\Scripts\00000000-SqlCmdSupport\00-CommandsAndVariables.sql`, I want the Script name to be `Scripts\00000000-SqlCmdSupport\00-CommandsAndVariables.sql`. 

Essentially, I want avoid putting the full path to the file in `SqlScript.Name` so that run-time differences from on DbUp upgrade to the next don't cause scripts to be re-executed. It is even arguable that the base folder name, i.e., `Scripts` should _**not**_ be included.

dazinator: Having changed the namespace of a project with embedded dbup scripts in recently, finding measures to prevent re-execution of the scripts was a pain. I'd be all for any change that makes that problem go away ;-)

JakeGinnivan: Cool, thats what I thought. So is this a breaking change?

src/DbUp/Support/SqlServer/SqlCommandReader.cs
rickculpepper: Yeah. I noticed that. We need a separation of concerns -- and namespaces! -- here.

Suggestions?

JakeGinnivan: Model after https://github.com/DbUp/DbUp/blob/master/src/DbUp.MySql/MySqlCommandReader.cs

Create a `MSSqlCommandReader` then use that instead of the SqlCommandReader

src/Samples/SqlServer.FolderRecursion/Program.cs
rickculpepper: Hmmm... I started to do this, but...

The SQLCMD stuff is pretty easy to add to the existing `SampleApplication`, although this should be SQL-Server-specific functionality and that seems like it should not be in the generic sample app. Just sayin'..

The file folder recursion is a whole different matter. I am pretty firm in my conviction that it needs to be clearly demonstrated and I question whether the docs are the right place to do that. Furthermore, we can't mix it in when the resource scripts in `SampleApplication` because the two are mutually exclusive. Of course, this is a database-independent features so, maybe it shouldn't be in a SQL-Server-specific sample application.

There. Now I've made an argument for not one, but *two* new sample apps. :P

We should discuss.

JakeGinnivan: At a minimum I think we need to add to the docs *as well*. Leave the samples for the moment but we need to add to docs and maybe link to the samples from the docs to make them discoverable

src/DbUp/Engine/Preprocessors/StripSchemaPreprocessor.cs
rickculpepper: Too late. ;) 

Actually, I like the succinctness of a single Regex expression. I’m going to do this… and I’ll do the same thing in `VariableSubstitutionPreprocessor.cs`

src/DbUp/ScriptProviders/FileSystemScriptProvider.cs
rickculpepper: @dazinator, we're on the same page here... ;)

@JakeGinnivan, Yes. This is a breaking change, but only for databases that have been previously upgraded `.WithScriptsFromFileSystem`

src/DbUp/Engine/Preprocessors/StripSchemaPreprocessor.cs
JakeGinnivan: The only downside to `\$\(?schema[\$\)]\.` is that this will match `$(schema$`. For this usecase, it's not really an issue I don't think, but may be an issue in the variable substitution preprocessor

src/DbUp/ScriptProviders/FileSystemScriptProvider.cs
JakeGinnivan: We need to add a migration in I think.. We could dynamically inject a script at the start of the script lists which updated the data in the schema versions table.

Not sure that this is a breaking change we can just introduce, the chance of double running migrations is too high. We also need to add tests around the upgrade scenario

src/DbUp/Support/SqlServer/SqlCommandReader.cs
rickculpepper: Okay. Can do. Will do, but know that I have concerns.
* If `SqlCommandReader` is going to remain database-independent, it should be moved out of the `DbUp.At some point, SqlCommandReader needs to be moved out of the `DbUp.Support.SqlServer` namespace.
* All the code related to SQL Server has has a prefix of `Sql...` up to this point. Diverging to use a prefix of `MSSql...` is a code smell as far as I'm concerned. :S

src/Samples/SqlServer.FolderRecursion/Program.cs
rickculpepper: Okay. I’ll look at this and see what I can come up with.

----------------------------------------------------------------------
NOTE:  The information in this email is confidential and may be legally 
privileged.  If you are not the intended recipient, we request that you 
(i) not read, use or disseminate the information, (ii) advise the sender 
immediately by reply email and (iii) delete this message and any 
attachments without retaining a copy.  Although this email and any 
attachments are believed to be free of any virus or other defect that 
may affect any computer system into which it is received and opened, it 
is the responsibility of the recipient to ensure that it is virus free 
and no responsibility is accepted by FreedomRoads, LLC or any of its 
affiliates for any loss or damage arising in any way from its use.

src/DbUp/Engine/Preprocessors/StripSchemaPreprocessor.cs
rickculpepper: Good point. Let me go fiddle with this some more.

----------------------------------------------------------------------
NOTE:  The information in this email is confidential and may be legally 
privileged.  If you are not the intended recipient, we request that you 
(i) not read, use or disseminate the information, (ii) advise the sender 
immediately by reply email and (iii) delete this message and any 
attachments without retaining a copy.  Although this email and any 
attachments are believed to be free of any virus or other defect that 
may affect any computer system into which it is received and opened, it 
is the responsibility of the recipient to ensure that it is virus free 
and no responsibility is accepted by FreedomRoads, LLC or any of its 
affiliates for any loss or damage arising in any way from its use.

src/DbUp/Support/SqlServer/SqlCommandReader.cs
JakeGinnivan: I have a branch locally which breaks MSSql support out of the core. But it is a breaking change.

We can leave this PR until I get the v4 branch underway then we can fix it properly, otherwise it is a decent temporary compromise

src/DbUp/ScriptProviders/FileSystemScriptProvider.cs
rickculpepper: Good point… but ouch! ;) The dynamic upgrade script seems like a good idea at first, but realistically how do we manage that? There are *so* many possibilities. This seems like a (very) slippery slope.

On the other hand I do understand the risk that this introduces. Let me go away and think about this a bit.

----------------------------------------------------------------------
NOTE:  The information in this email is confidential and may be legally 
privileged.  If you are not the intended recipient, we request that you 
(i) not read, use or disseminate the information, (ii) advise the sender 
immediately by reply email and (iii) delete this message and any 
attachments without retaining a copy.  Although this email and any 
attachments are believed to be free of any virus or other defect that 
may affect any computer system into which it is received and opened, it 
is the responsibility of the recipient to ensure that it is virus free 
and no responsibility is accepted by FreedomRoads, LLC or any of its 
affiliates for any loss or damage arising in any way from its use.

src/DbUp/Support/SqlServer/SqlCommandReader.cs
rickculpepper: Let’s hold off for v4. No one is screaming for these features right now. I’d *like* to have them, but I can wait…

----------------------------------------------------------------------
NOTE:  The information in this email is confidential and may be legally 
privileged.  If you are not the intended recipient, we request that you 
(i) not read, use or disseminate the information, (ii) advise the sender 
immediately by reply email and (iii) delete this message and any 
attachments without retaining a copy.  Although this email and any 
attachments are believed to be free of any virus or other defect that 
may affect any computer system into which it is received and opened, it 
is the responsibility of the recipient to ensure that it is virus free 
and no responsibility is accepted by FreedomRoads, LLC or any of its 
affiliates for any loss or damage arising in any way from its use.

rickculpepper: I see the 4.0.0 branch in DbUp. Should I look at that?

----------------------------------------------------------------------
NOTE:  The information in this email is confidential and may be legally 
privileged.  If you are not the intended recipient, we request that you 
(i) not read, use or disseminate the information, (ii) advise the sender 
immediately by reply email and (iii) delete this message and any 
attachments without retaining a copy.  Although this email and any 
attachments are believed to be free of any virus or other defect that 
may affect any computer system into which it is received and opened, it 
is the responsibility of the recipient to ensure that it is virus free 
and no responsibility is accepted by FreedomRoads, LLC or any of its 
affiliates for any loss or damage arising in any way from its use.

JakeGinnivan: Have not got far, but you can have a look at https://github.com/DbUp/DbUp/pull/130 which shows where I am going

rickculpepper: Will do!

----------------------------------------------------------------------
NOTE:  The information in this email is confidential and may be legally 
privileged.  If you are not the intended recipient, we request that you 
(i) not read, use or disseminate the information, (ii) advise the sender 
immediately by reply email and (iii) delete this message and any 
attachments without retaining a copy.  Although this email and any 
attachments are believed to be free of any virus or other defect that 
may affect any computer system into which it is received and opened, it 
is the responsibility of the recipient to ensure that it is virus free 
and no responsibility is accepted by FreedomRoads, LLC or any of its 
affiliates for any loss or damage arising in any way from its use.

src/DbUp/ScriptProviders/FileSystemScriptProvider.cs
dazinator: Tricky one! When dbup runs to check if the schema table exists, it could migrate it if necessary at that point i guess.

Would calculating some kind of hash for the script and having that as a column in the schemaversions table be a good idea? In theory that would allow us to prevent running a duplicate script even if you changed its name!

JakeGinnivan: We can use the same approach, we never change the original schemaversions table, then create change scripts. They will be run before user scripts.

rickculpepper: @dazinator - I've been ruminating on a way to know that a script "matches" and a hash never occurred to me... DOH! I think this is  a great idea...

@JakeGinnivan - Using a change script seems like a really good choice here... DbUp should [eat it's own dog food](https://en.wikipedia.org/wiki/Eating_your_own_dog_food)! How about the original SchemaVersions script and the change script both become resources in the DbUp.dll? We'll use VariableReplacement to continue to allow the caller to specify the name of the SchemaVersions table... I'm liking this... A LOT!

A couple of issues come to mind. This is just FYI. Not asking for answers here. I'll go see if I can work these out
* How to insert these scripts into the very tip-top of the script execution sequence
* Journaling for DbUp scripts vs. user scripts.
* Assuming that this is (for the present) about adding the checksum that @dazinator  mentioned, it seems like `DbUp.Engine.UpgradeEngine` might need a new method that would calculate the checksum for previously executed scripts.
```C#
public DatabaseUpgradeResult UpgradeChecksums(string latestScript){...}
```

dazinator: I like the hash idea in principal but it introduces some caveats. Like
possible hash collisions (2 different scripts with idwntical hashes?
although guessing this would be unlikely) and what about if someone had 2
identical migration scripts but with different filenames because 1 needs to
be run first and another again later during a migration.. Its an edge case
but if the hash was based on script content alone then it potentially
wouldnt execute the second script... so i like the idea in principal but
perhaps thinking it would need to be an "opt in" feature in the api. So
dbup by all means could calculate and populate hashes for scripts, but it
just wouldnt enforce hash checks pre executing a script unless you opted in
via the api somehow?

JakeGinnivan:
@ternaryalchemy are you able to break this up into a number of smaller PR's against the release/4.0.0 branch?

This is just too big for me to understand what is happening and get it in with confidence.

Shogan:
Hi, I know this is closed, and looking at the detail, it seems that some work is needed to split this feature up into smaller pull request chunks, but I'd just like to chime in that I would love to have SQLCMD support with DbUp.

PR #137, opened 2015-09-20
Multiple Fixes, and Split SqlServer Out

### LINKED ISSUE #83: SqlScriptExecutor should be pulled into core
Several database providers (Postgres, MySQL and Sqlite) use the SqlScriptExecutor which is currently under the SQL Server namespace. This results in log messages being generated with the words **SQL Server** in them:

```
Executing SQL Server script 'App.Data.Up.Scripts.Script-0001.sql'
```

If there is appetite for the change I'll send a PR.

Thoughts?

### LINKED ISSUE #134: Pull SqlServer out of the Core
Each provider has it's own project, except ms sql server..
We should pull this out into its own provider

### LINKED ISSUE #135: Provider Exception Handling
Most of the providers rely on the `SqlScriptExecutor` which is MS SqlServer oriented - if there are exceptions when a script is executed they are not handled and logged as provider specific exception - i.e for Firebird they are not handled as `FbException` - instead they are handled as `DbException` and only for SqlServer are they handled / logged as the more specific `SqlException`.

This can be improved, introduce an abstract base class `ScriptExecutor` and have provider specific implmentations such as `SqlServerScriptExecutor`/ `FirebirdScriptExecutor` etc. Those implementations can handle exceptions in a mannor concrete to the provider. This may mean a bit more information can be output in the logs when things go wrong.

### LINKED ISSUE #136: Duplicate Schema Table Exists Check
The Table Journal checks twice for the existence of the Schema Table (by doing a select count) - once should be enough.

For example, from an approval test

```
Open connection
Begin transaction
Execute scalar command: select count(*) from [SchemaVersions]
Dispose command
Commit transaction
Dispose transaction
Begin transaction
Execute non query command: print 'script1'
Dispose command
Commit transaction
Dispose transaction
Begin transaction
Execute scalar command: select count(*) from [SchemaVersions]
Dispose command
Commit transaction
Dispose transaction
Begin transaction
Execute non query command: create table [SchemaVersions] (
    [Id] int identity(1,1) not null constraint PK_SchemaVersions_Id primary key,
    [ScriptName] nvarchar(255) not null,
    [Applied] datetime not null
)

```

dazinator:
Fixes #134 - Move SqlServer into it's own provider.
Fixes #136 - Duplicate checks for whether Schema Table exists.
Fixes #135 and Fixes #83 - Better exception handling for all non MS providers.


src/DbUp.Tests/DatabaseSupportTests.VerifyBasicSupport.SQLite.approved.txt
```diff
@@ -1,11 +1,9 @@
 ﻿Open connection
-Execute scalar command: select count(*) from [SchemaVersions]
+Execute scalar command: select count(*) from SchemaVersions
```
JakeGinnivan: Any reason for dropping the `[` and `]`?

src/DbUp.Tests/DbUp.Firebird.approved.cs
```diff
@@ -16,6 +16,21 @@ public class FirebirdPreprocessor : DbUp.Engine.IScriptPreprocessor
...
         public string Process(string contents) { }
     }
+    public class FirebirdScriptExecutor : DbUp.Support.ScriptExecutor
+    {
+        public FirebirdScriptExecutor(System.Func<DbUp.Engine.Transactions.IConnectionManager> connectionManagerFactory, System.Func<DbUp.Engine.Output.IUpgradeLog> log, string schema, System.Func<bool> variablesEnabled, System.Collections.Generic.IEnumerable<DbUp.Engine.IScriptPreprocessor> scriptPreprocessors) { }
+        protected override void ExecuteCommandsWithinExceptionHandler(int index, DbUp.Engine.SqlScript script, System.Action excuteCommand) { }
+        protected override string GetVerifySchemaSql(string schema) { }
+        protected override string QuoteSqlObjectName(string objectName) { }
+    }
+    public sealed class FirebirdTableJournal : DbUp.Support.TableJournal
```
JakeGinnivan: I don't think we need to seal

src/DbUp.Tests/TransactionScenarios.UsingSingleTransactionScenarioSuccess.approved.txt
```diff
@@ -4,13 +4,11 @@ Execute scalar command: select count(*) from [SchemaVersions]
...
 Execute non query command: print 'script1'
 Dispose command
-Execute scalar command: select count(*) from [SchemaVersions]
-Dispose command
 Execute non query command: create table [SchemaVersions] (
-	[Id] int identity(1,1) not null constraint PK_SchemaVersions_Id primary key,
-	[ScriptName] nvarchar(255) not null,
-	[Applied] datetime not null
-)
+        	[Id] int identity(1,1) not null constraint PK_SchemaVersions_Id primary key,
```
JakeGinnivan: Just for the sake of the diffs, would be good to fix the indenting of these guys

src/DbUp.sln
```diff
@@ -37,60 +37,118 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DbUp.MySql", "DbUp.MySql\Db
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DbUp.SQLite.Mono", "DbUp.SQLite.Mono\DbUp.SQLite.Mono.csproj", "{5C474B5C-019C-4C03-9E20-2D0CDEAAEA7B}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DbUp.SqlServer", "DbUp.SqlServer\DbUp.SqlServer.csproj", "{EE71E015-3442-45B3-97B6-7886B7DE8351}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
```
JakeGinnivan: Can we remove the Any CPU target and just leave the mixed platforms one (I know VS automatically puts it in when you add projects.

JakeGinnivan:
This looks good in general. Few things

 - In my PR I have created DbUp.Core and make DbUp an obsolete package, see https://github.com/DbUp/DbUp/pull/130/files#diff-97b746221eb1d9b7ff383a0c8ae39807R5
   - This is so when people upgrade from DbUp it will automatically bring in the SqlServer package. I plan on putting a readme so people can then see that if they are not using SqlServer then they can uninstall DbUp and the SqlServer package to leave their target database support and core
 - I think we may be overusing base classes/inheritance. See https://github.com/DbUp/DbUp/commit/1f38138670104a2f5c958184ab70d434e25bc6ef - rather than overrides I am passing a factory into the base, it is a much more flexible approach. I would like to see us put some additional thought about where we could use composition to have a single Journal implementation and pass a dependency into it rather than an abstract journal with overrides

src/DbUp.Tests/DatabaseSupportTests.VerifyBasicSupport.SQLite.approved.txt
dazinator: Nope! Let me fix that

src/DbUp.Tests/TransactionScenarios.UsingSingleTransactionScenarioSuccess.approved.txt
dazinator: Ok will look at that!

src/DbUp.sln
dazinator: Ok!

src/DbUp.Tests/DbUp.Firebird.approved.cs
dazinator: Ok. Think the original FirebirdTableJournal was sealed so just kept it. Will remove from all.

dazinator:
Ok - I will have a go at fixing up to produce the same upgrade path as mentioned.

I agree composition over inheritance is better as a general approach. Scrapping inheritance for the tablejournals and using composition instead should result in some more specific services that need to be implemented per provider instead. I think that a move to composition can best be addressed by a PR focused on that concern, after we have the providers split out correctly.. I dont really want to conflate this PR any further.

Also happy for you to merge your PR and lose this one if thats easier - but would be nice if we can keep the fixes around redundant table exists checks and the slightly improved logging!

If we do compose the TableJournals then I can see it needing several provider specific services:

- it needs to allow the dbcommands to be customised / set by the provider (so the provider gets to set command text and params etc) 
- it needs to be able to execute dbcommands with provider specific exceptions handled (ScriptExecutor has this logic but it could be refactored out into its own service for use by TableJournal also)
- it needs to be able to quote identifiers appropriately. 
- it needs to allow providers to hook into after the journal table is created so that the provider can do further customisations.(see FirebirdTableJournal it has a few)

.. and maybe other things :)

JakeGinnivan:
> it needs to allow the dbcommands to be customised / set by the provider (so the provider gets to set command text and params etc)

We have the DbCommandFactory for this at the moment. .NET abstracts this for us

> it needs to be able to execute dbcommands with provider specific exceptions handled (ScriptExecutor has this logic but it could be refactored out into its own service for use by TableJournal also)

DatabaseManager (which is per database) could be responsible for handling exceptions. Maybe a method with signature `bool HandleException(Exception ex)`, if it returns false it logs generically and rethrows or whatever the current behaviour is. 

Just ideas, when putting in the MySql parser changes the inheritance approach blocked me which is why I had to change.

dazinator:
> We have the DbCommandFactory for this at the moment. .NET abstracts this for us

This helps new up an instance of an appropriate DbCommand, but we need the provider specific command text to be set on that command prior to execution - as the SQL syntax is not identical accross all providers! No biggie - just an added consideration that require a little more thought.

> DatabaseManager (which is per database) could be responsible for handling exceptions. Maybe a method with signature bool HandleException(Exception ex), if it returns false it logs generically and rethrows or whatever the current behaviour is.

Yeah - agree we need something like that.. Think i prefer the approach within this PR though but obviously I am a bit biased there - where execution of the command happens within a provider specific wrapper / delegate that contains providers try / catch.

I'll have a go at addressing some of this stuff at the weekend! Let me know if it's looking like any of this PR is starting to become obsolete for any reason though and I won't waste any further time on it ;)

dazinator:
>Just ideas, when putting in the MySql parser changes the inheritance approach blocked me which is why I had to change.

Whats this MySql parser you speak of!?

JakeGinnivan:
https://github.com/DbUp/DbUp/blob/master/src/DbUp.MySql/MySqlCommandReader.cs

dazinator:
Tidied up based on your feedback. I'm not sure I dealt with the AnyCpu change you asked for correctly..
Also I haven't tried to address the NuGet upgrade path yet to make it as per your suggestion.

PR #143, opened 2015-11-04
Docs/convention check

teyc:
Here's a little powershell script that adds an Embedded Resource convention check to DbUp projects.. Usage: In Package Console 

`> Add-DbUpEmbeddedCheck`


JakeGinnivan:
This is quite cool, thanks!

PR #150, opened 2015-12-02
Alter Journal Table Creation To Uniquely Name Primary Key

### LINKED ISSUE #146: Naming of Journal Primary Key Constraint - Prevents more than one Migrator
For each of the Journals (FireBirdTableJournal, PostgresqlTableJournal, SqLiteTableJournal and SqlTableJournal), the creation of the Journal Table is done something like below:

`
         protected virtual string CreateTableSql(string tableName)
        {
            return string.Format(@"create table {0} (
	[Id] int identity(1,1) not null constraint PK_SchemaVersions_Id primary key,
	[ScriptName] nvarchar(255) not null,
	[Applied] datetime not null
)", tableName);
        }
`

Note that the constraint is always called "PK_SchemaVersions_Id", no matter what the table name is. So if you try and create a second Journal Table, it will fail.

This means that if you (1) Have two migrators using two different table names, or (2) change the table name during development after running the migration once, all future migrations will fail.

I am unsure of the best solution for this long term is. Simply adding the table name to the constraint name will not work as it is the full escaped name (eg [dbo].[Journal Table Name]).

AdrianJSClark:
Changes the "CREATE TABLE" scripts to ensure that named primary key constraints include the given table name. Avoiding name collisions if two differently-named journal tables are created within a single database.

Impacts the following Journals:
 - SQL Server and SQL CE
 - PostgreSQL
 - Firebird

Fixes #146

### DIFF
--- a/src/DbUp.Tests/DatabaseSupportTests.VerifyBasicSupport.SqlCe.approved.txt
+++ b/src/DbUp.Tests/DatabaseSupportTests.VerifyBasicSupport.SqlCe.approved.txt
@@ -6,7 +6,7 @@ Dispose command
 Execute scalar command: select count(*) from [SchemaVersions]
 Dispose command
 Execute non query command: create table [SchemaVersions] (
-	[Id] int identity(1,1) not null constraint PK_SchemaVersions_Id primary key,
+	[Id] int identity(1,1) not null constraint [PK_SchemaVersions_Id] primary key,
 	[ScriptName] nvarchar(255) not null,
 	[Applied] datetime not null
 )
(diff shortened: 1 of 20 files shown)


bingnz:
+1: this would be useful!

floradu88:
+1: I wanted to submit the same pull request. Could the owner/maintainers please add this to the next release?

JakeGinnivan:
Will try to get this in and released this weekend

AdrianJSClark:
@JakeGinnivan Get in touch if you need any help or any changes made here.

floradu88:
Thanks @AdrianJSClark and @JakeGinnivan

bingnz:
Thanks @AdrianJSClark  and @JakeGinnivan. There seems to be a bug in the SQL table journal though; if you specify a schema name it attempts to prepend it to the primary key name which is invalid syntax. The code in `SqlTableJournal` should be:

    protected virtual string CreatePrimaryKeyName(string schema, string table)
    {
        return SqlObjectParser.QuoteSqlObjectName("PK_" + table + "_Id");
    }

AdrianJSClark:
Oops! My bad. Thanks for pointing that out @bingnz. The latest version of this branch should correct that error.

PR #162, opened 2016-02-02
Remove reliance on exceptions to check if table exists (fixes #142)

### LINKED ISSUE #142: Version 3.3.0 doesn't create SchemaVersions table
Upon PerformUpgrade() I get the following message: Invalid object name 'SchemaVersions'.

The issue is not present in version 3.2.1. 
A change wasn't announced in the release notes, so I consider this as a bug.

omnibs:
The code that checks if a table exists was relying on doing a `select count(*) from {table}` and expecting an Exception from the database when the table was not there.

It's not clear to me why showing the Exception while debugging, as seen on #142, altered the flow of execution, or how this could be behaving differently in 3.2.1, but since relying on exceptions for flow control is smelly I've implemented db-specific checks for SqlServer and SQLite.

For that I needed to change the public API and the expected log files so this PR grew a lil bit.

### DIFF
--- a/src/DbUp/Support/SqlServer/SqlTableJournal.cs
+++ b/src/DbUp/Support/SqlServer/SqlTableJournal.cs
@@ -171,10 +171,7 @@ namespace DbUp.Support.SqlServer
                 {
                     using (var command = dbCommandFactory())
                     {
-                        command.CommandText = string.Format("select count(*) from {0}", CreateTableName(schema, table));
-                        command.CommandType = CommandType.Text;
-                        command.ExecuteScalar();
-                        return true;
+                        return VerifyTableExistsCommand(command, table, schema);
                     }
                 }
                 catch (SqlException)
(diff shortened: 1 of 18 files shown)


src/DbUp/Support/SqlServer/SqlTableJournal.cs
```diff
@@ -171,10 +171,7 @@ private bool DoesTableExist()
...
                     using (var command = dbCommandFactory())
                     {
-                        command.CommandText = string.Format("select count(*) from {0}", CreateTableName(schema, table));
-                        command.CommandType = CommandType.Text;
-                        command.ExecuteScalar();
-                        return true;
+                        return VerifyTableExistsCommand(command, table, schema);
                     }
                 }
                 catch (SqlException)
```
markoueis: With these changes, do we still need the exceptions here. They might hurt. See issue #271

PR #165, opened 2016-02-08
Split sql server out

JakeGinnivan:
Based on #137

@dazinator have rebased your PR, tried to re-include the bug fixes and such. Could you have a bit of a look and see what you think?


dazinator:
Thanks Jake, I'll take look!

dazinator:
Thanks for all this work rebasing this!
In general I think it's all there, but it does seem to have some failing tests at present:-

image

Would be great if we could get these tests running automatically on AppVeyor!

src/DbUp/Support/TableJournal.cs
```diff
@@ -0,0 +1,209 @@
...
+
+            journalExists = true;
+        }
+
+        protected bool DoesTableExist()
+        {
+            Log().WriteInformation("Checking whether journal table exists..");
+            return ConnectionManager().ExecuteCommandsWithManagedConnection(dbCommandFactory =>
+            {
+                using (var command = dbCommandFactory())
```
dazinator: Need to add back the Try / Catch around this, as per: https://github.com/dazinator/DbUp/commit/442a8eda71ae3727749684b9c1f6a1ec91c26f96#diff-ce8df5609e9b4c5d25e0237203ba7205R183 - this should failing test

src/DbUp.SQLite/SQLiteTableJournal.cs
```diff
@@ -0,0 +1,47 @@
...
+$@"CREATE TABLE {FqSchemaTableName} (
+    SchemaVersionID INTEGER CONSTRAINT {quotedPrimaryKeyName} PRIMARY KEY AUTOINCREMENT NOT NULL,
+    ScriptName TEXT NOT NULL,
+    Applied DATETIME NOT NULL
+)";
+        }
+
+        protected override string DoesTableExistSql()
+        {
+            return $"SELECT count(*) FROM sqlite_master WHERE type = 'table' AND name = '{UnquotedSchemaTableName}' COLLATE NOCASE";
```
dazinator: When this sql is executed during this test, against the in memory sql lite database:
image
The result of the ExecuteScalar() is null. Therefore the test fails because it doesn't detect the table was created, even after it creates it.

JakeGinnivan:
@dazinator ugh sorry

nCrunch is not running all tests. Will fix the fails :P

src/DbUp/Support/TableJournal.cs
JakeGinnivan: This one is due to we actually don't throw exceptions. Fixed by returning 0 instead

src/DbUp.SQLite/SQLiteTableJournal.cs
JakeGinnivan: Ugh, no idea about this one.. going to leave for tonight.

I think I might get rid of these sort of tests then do manual integration tests when I do large changes like this. If you have ideas about this or want to have a crack at fixing, be my guest :P

PR #77, opened 2014-12-11
Oracle support for DbUP

SimonRupar:
Based on pull-request #53 

Done: 
1. Added Oracle provider for DbUp. Because Oracle does not support roll-back transaction, all scripts have hash of content. More on pull-reqest #53 under point number 6.
2. Added console application for oracle. 
3. Added unit-tests for testing oracle support provider.

PR #171, opened 2016-02-17
Case sensitive INFORMATION_SCHEMA table in collations other than Latin

### LINKED ISSUE #170: SqlTableJournal could not find SchemaVersion table if the database collation is not set to Latin*
Hi,
SqlTableJournal could not find SchemaVersion table if the database collation is not set to Latin* because of that DbUp is always executing all scripts.

Solution; the table name( information_schema.tables) must be written upper-case  in the lookup query. https://github.com/DbUp/DbUp/blob/master/src/DbUp/Support/SqlServer/SqlTableJournal.cs#L195-L197
```cs
command.CommandText = string.IsNullOrEmpty(schema)
  ? string.Format("select 1 from information_schema.tables where TABLE_NAME = '{0}'", tableName)
  : string.Format("select 1 from information_schema.tables where TABLE_NAME = '{0}' and TABLE_SCHEMA = '{1}'", tableName, schemaName);
```

omnibs:
This fixes #170

### DIFF
--- a/src/DbUp.Tests/DatabaseSupportTests.VerifyBasicSupport.Firebird.approved.txt
+++ b/src/DbUp.Tests/DatabaseSupportTests.VerifyBasicSupport.Firebird.approved.txt
@@ -1,9 +1,9 @@
 ﻿Open connection
-Execute scalar command: select count(*) from schemaversions
+Execute scalar command: select 1 from RDB$RELATIONS where RDB$SYSTEM_FLAG = 0 and RDB$RELATION_NAME = 'schemaversions'
 Dispose command
 Execute non query command: script1contents
 Dispose command
-Execute scalar command: select count(*) from schemaversions
+Execute scalar command: select 1 from RDB$RELATIONS where RDB$SYSTEM_FLAG = 0 and RDB$RELATION_NAME = 'schemaversions'
 Dispose command
 Execute non query command: CREATE TABLE schemaversions
                                     (
--- a/src/DbUp.Tests/DatabaseSupportTests.VerifyBasicSupport.MySql.approved.txt
+++ b/src/DbUp.Tests/DatabaseSupportTests.VerifyBasicSupport.MySql.approved.txt
@@ -1,9 +1,9 @@
 ﻿Open connection
-Execute scalar command: select count(*) from `schemaversions`
+Execute scalar command: select 1 from INFORMATION_SCHEMA.TABLES where TABLE_NAME = 'schemaversions'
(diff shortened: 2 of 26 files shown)


JakeGinnivan:
Does this issue affect other databases?

In the 4.0 branch I was actually thinking about making this SQL standard for all databases (as information_schema is standard not a SQL server thing I think). I.e moving this down into the Journal base class and then only overriding if needed. Thoughts?

Either way, this fix needs to be applied to both master and release/4.0.0. Either by forward merging master into release/4.0.0 after this is merged or manually.

omnibs:
Possibly, haven't tested. Would require some research.

Seems like a good idea to make it standard, I did google for a standard way to check tables and information_schema is what came out.

I'll handle that on 4.0.0 too.

JakeGinnivan:
What is the status of this?

omnibs:
I hadn't done my homework.

(According to wikipedia)[https://en.wikipedia.org/wiki/Information_schema#Implementation] INFORMATION_SCHEMA is supported by:

* Microsoft SQL Server
* MySQL
* PostgreSQL

And is unsupported by Oracle and Sqlite.

Since Oracle support is still a PR@#77, I believe it's OK to make INFORMATION_SCHEMA an overridable default.

I'm gonna work on this right now and report back by updating the PR.

About 4.0, I'll rewrite the changes on that branch and try to PR to it instead of master, I'm not sure it's possible but a quick google shows it might be.

omnibs:
[Apparently Firebase does not support INFORMATION_SCHEMA](http://www.alberton.info/firebird_sql_meta_info.html#.Vt2NEpwrKUk), so I used these weird tables, but I don't have a Firebird DB to test it.

On the MySql tests I got some different flow going on `src\DbUp.Tests\Support\MySql\MySqlSupportTests.CanHandleDelimiter.approved.`. On your test the Journal table was taken as present, on my test it's taken as nonexistant, so it's creating the table on line 21. Maybe it's more correct like this, since we're simulating an empty DB on the test. What you think?

JakeGinnivan:
> About 4.0, I'll rewrite the changes on that branch and try to PR to it instead of master, I'm not sure it's possible but a quick google shows it might be.

git rebase upstream/release/4.0.0

That will rewrite your history onto the 4.0.0 branch.

JakeGinnivan:
This pr will look quite different on the 4.0 branch. We use schema versions already for most DB's. I think it just needs to be pushed into the base class by default then overriden by the ones that don't support it.

PR #173, opened 2016-03-01
Refactored StoreExecutedScript for extensibility

Ben-Farrimond:
I found myself wanting to add extra columns to the SchemaVersions table to store additional information - fortunately there were extensibility points present in this class to add additional columns to the table. However, when I came to insert into it, the StoreExecutedScript method was not virtual and the insert script and db parameters were embedded inside of this method.

All I've done is simply move them out and make them 'protected virtual' to allow for inserting into columns that others may want to add.

### DIFF
--- a/src/DbUp/Support/SqlServer/SqlTableJournal.cs
+++ b/src/DbUp/Support/SqlServer/SqlTableJournal.cs
@@ -110,25 +110,46 @@ namespace DbUp.Support.SqlServer
             {
                 using (var command = dbCommandFactory())
                 {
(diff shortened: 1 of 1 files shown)


JakeGinnivan:
Hey @Ben-Farrimond 

Could you resubmit this PR against the `release/4.0.0` branch. The journal table code is quite different on that branch and this is a useful change.

JakeGinnivan:
Closed because I don't want features in the v3 branch without being in v4

PR #175, opened 2016-03-04
Fix for running dbup on Azure boxes.

ammaraltahan:
I could increase the timeout through a configure like the example in this article "https://phejndorf.wordpress.com/2014/07/02/setting-execution-timeout-in-dbup/" but unfortunately this is not possible for EnsureDatabase.For.SqlServer() method.

### DIFF
--- a/src/DbUp/Support/SqlServer/SqlServerExtensions.cs
+++ b/src/DbUp/Support/SqlServer/SqlServerExtensions.cs
@@ -185,7 +185,8 @@ public static class SqlServerExtensions
             // Create the database...
             using (var command = new SqlCommand(sqlCommandText, connection)
             {
-                CommandType = CommandType.Text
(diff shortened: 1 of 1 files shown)


JakeGinnivan:
Could this be made a configuration setting?

ammaraltahan:
Actually I ended up adding an extension method with the same code but with
increased timeout.

As for your question, it is certainly a good solution but requires more
complex changes.

JakeGinnivan:
@ammaraltahan I mean, would you be willing to make this a proper configuration setting to make it easier for others?

ammaraltahan:
Sure!

JakeGinnivan:
Awesome, thanks!

JakeGinnivan:
Superseded by #179

PR #178, opened 2016-03-09
Exposed a new overload to take command timeout

ammaraltahan:
With this overload, if create database statement times out (especially in Azure databases), increase the time out by using the added overload. Otherwise the value is set to -1 and it won't be set unless it is greater than or equal zero.

Edit---
Oops I forgot to remove "private static int _createDbCommandTimeout = 0;" and it's there by mistake, what do I do? create a new pull request?


ammaraltahan:
This submission has a mistake, please delete this one, I'm not sure how to retract/delete my submission?!

PR #183, opened 2016-04-05
Changed cast to convert to support longs

chriswithpants:
My local MySQL install is returning `object{long}` when the `select 1...` executes when checking for the journal table. That long won't cast `as int?`.


mgkeen:
I noticed exactly the same thing this morning and you just beat me to it with the PR!

PR #185, opened 2016-04-05
Provide overloads to specify MySql schema to use in extensions

mgkeen:
When multiple schemas existed, it checked for any table named schemaversions in any schema.
Query code was already in place to be specific about schema, but not exposed via extensions.

### DIFF
--- a/src/DbUp.MySql/MySqlExtensions.cs
+++ b/src/DbUp.MySql/MySqlExtensions.cs
@@ -28,16 +28,43 @@ public static class MySqlExtensions
     /// <summary>
     /// Creates an upgrader for MySql databases.
     /// </summary>
+    /// <param name="supported">Fluent helper type.</param>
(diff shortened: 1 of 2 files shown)


chriswithpants:
Build failed :disappointed:

mgkeen:
Build failure appears to be because i added to the public API and it needs approval. Don't want to go and change the "approved" file myself. I assume @JakeGinnivan needs to do this?

JakeGinnivan:
@mgkeen you can approve no worries.

The check is there so if you change the public API you can visualise that change and make sure it is not a breaking change.

JakeGinnivan:
This change is just adding overloads, so it's fine to accept the change because there are no breaking changes.

mgkeen:
Updated. Really like that way of visualising API changes. Been looking for something that simple.

JakeGinnivan:
https://github.com/JakeGinnivan/ApiApprover is the project :)

PR #187, opened 2016-04-07
Target new VersionFoldersScriptsProvider to branch Release/4.0.0

rintje:
Merged readme.md docs for VersionFoldersScriptProvider into docs/more-info/script-providers.md

### DIFF
--- a/docs/more-info/script-providers.md
+++ b/docs/more-info/script-providers.md
@@ -22,11 +22,6 @@ builder.WithScriptsEmbeddedInAssemblies(new[]
 (string s) => s.StartsWith("Script"))
 ```
 
-## FileSystemScriptProvider
-Finds scripts in a specified directory
-### Usage
-`builder.WithScriptsFromFileSystem(path)`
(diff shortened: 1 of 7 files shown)


rintje:
OK, finally got the git stuff sorted :-)

JakeGinnivan:
I know this has been ages, but I wonder if we should separate the script provider and the script sorter. 

.ScriptsEmbeddedInAssembly(..)
.SortScriptsAlphanumerically() // .SortScriptsUsing<VersionFolderScriptSorter>()

Thoughts?

rintje:
Sounds like a good idea. That makes room for additional sort types like SemVer.

The VersionFolderScriptProvider currently also filters scripts using the Version class, to omit scripts with a higher version than the version specified in param targetVersion. Thus, if sorting is made a separate functionality, then i guess filtering should also be separated? For example: FilterScriptsByVersionUsing()

rintje:
note: see #231 first

Issue #189, opened 2016-04-27
OutOfMemoryException in variable substitution


ammaraltahan:
I have a large file (data inserts) that has one variable at the top of the file (i.e.Using [$Database_Name$]) and it errors out with this exception.

Is this a limitation for Regex? or the regex is too greedy? may be worth looking into.

System.OutOfMemoryException: Exception of type 'System.OutOfMemoryException' was
 thrown.
   at System.Text.RegularExpressions.RegexRunner.DoubleTrack()
   at System.Text.RegularExpressions.RegexRunner.EnsureStorage()
   at System.Text.RegularExpressions.RegexInterpreter.Goto(Int32 newpos)
   at System.Text.RegularExpressions.RegexInterpreter.Go()
   at System.Text.RegularExpressions.RegexRunner.Scan(Regex regex, String text,
Int32 textbeg, Int32 textend, Int32 textstart, Int32 prevlen, Boolean quick, Tim
eSpan timeout)
   at System.Text.RegularExpressions.Regex.Run(Boolean quick, Int32 prevlen, Str
ing input, Int32 beginning, Int32 length, Int32 startat)
   at System.Text.RegularExpressions.Regex.Match(String input, Int32 startat)
   at System.Text.RegularExpressions.Regex.Match(String input)
   at DbUp.Engine.Preprocessors.VariableSubstitutionPreprocessor.Process(String

omnibs:
What's the size of the file? I'll try to reproduce the issue and see what can be done.

ammaraltahan:
46.5 Mbs !! :)

omnibs:
That's not a lot! How much ram do you have? Are you running as a console app or embedded in an ASP.NET application?

ammaraltahan:
I have 16g RAM and yes I'm running as console app.

omnibs:
Just tried adding a 52mb script with lots of comments and random `select *` lines to the sample app in the repo. Ran just fine. 

I'd guess there's something specific to your script that's making the regex go crazy, like [this Atom regex bug](http://davidvgalbraith.com/how-i-fixed-atom/) that made the editor hang whenever enter key was pressed at the end of a very specific line of Go code.

My suspect is this huge regex in `VariableSubstitutionPreprocessor` to prevent variables from being substituted inside comments:

```c#
[...]

PR #190, opened 2016-04-28
Fix for #189 variable substitution OutOfMemory error

omnibs:
The regex to prevent throwing errors at unset variables inside `VariableSubstitutionPreprocessor` is too complex and susceptible to stack overflow and out of memory errors.

I've rewritten what it does without regex, parsing the comments and marking the index ranges, with tests and all.

One thing I've noticed writing this is that this method of dealing with comments is very error prone if scripts contain strings with `/*` or `--`. 

It's an issue in the current code and also in this new one I'm proposing. In order to handle that we'd have to parse strings in the query also. I'm willing to try that out in a second PR, since this is already an issue in the master branch.


dazinator:
Not positive if this is related but could the `SqlCommandReader` class that we have help here at all? It's basically a reader that is intelligent enough to read SQL files, with semantics for reading comment blocks (--, */) and splitting on GO statements that are not within comment blocks. We use it so that we can read 1 "command" at a time, even if the script contains multiple commands with dash dash or slash star comments. We used to use a regex for splitting on GO but it was pretty ugly imho and not very robust!

omnibs:
Good idea @dazinator, seems like the perfect candidate. 

I've seen we have one for SqlServer that apparently is also used for SqlCe, one for MySql. Do you have any idea what kind of trouble could we run into using that class for pgsql or firebird?

dazinator:
I had a quick review of the reader. Because both those systems use the sql ansi standard for comments format (single line and multiling) and for string literals, the bulk of the logic in the reader should work fine with postgres and firebird.

However one minor problem I did spot is that the reader currently assumes that [] are used to escape object names. The reader reads past bracketed values so it doesn't falsely detect things like "GO" within object names. This works fine for the MSSQL etc but Firebird uses double quotes, ". not [].

So basically, one tweak we'd probably have to make is to allow those characters to be configurable / overriden rather than hardcoded to [] and then that should be ok.

omnibs:
Work work work... finally had some time to get back to this issue.

I've taken a look at `SqlCommandReader`. In order to use it I'd have to change it in ways that make no sense for a `SqlCommandReader`. 

I'm trying my hand at breaking it apart. Extracting a `SqlParser`, implementing the original `SqlCommandReader` over it and porting variable substitution to that.

JakeGinnivan:
Hey guys, I am going to push up the netcore branch tonight. Do you recon you will have time over the next few days to look a the SqlCommandReader approach and get into the v4 branch?

omnibs:
Jake, thanks for the reminder, just made some time to come back and finish this.

I had a hard time figuring out why we needed to check [this](https://github.com/DbUp/DbUp/blob/master/src/DbUp/Support/SqlServer/SqlCommandReader.cs#L431-L434) on `ReadDelimiter`:

```C#
if (Read(buffer, 0, Delimiter.Length - 1) != Delimiter.Length - 1)
{
    return true;
}
```

When we check for delimiters we already peek ahead for the delimiter's length [here](https://github.com/DbUp/DbUp/blob/master/src/DbUp/Support/SqlServer/SqlCommandReader.cs#L336) on `IsBeginningOfDelimiter`, so I believe that `if` could never be true.

```C#
TryPeek(Delimiter.Length - 1, out result)
```

Could you guys double check I haven't missed anything?

Jake, I can squash these commits if you give me a thumbs up, I've just left them here for clarity.

src/DbUp/Engine/Preprocessors/VariableSubstitutionPreprocessor.cs
```diff
@@ -29,47 +33,82 @@ public VariableSubstitutionPreprocessor(IDictionary<string, string> variables)
         /// <param name="contents"></param>
         public string Process(string contents)
         {
-            Regex regexObj = new Regex(@"/\*(?>(?:(?!\*/|/\*).)*)(?>(?:/\*(?>(?:(?!\*/|/\*).)*)\*/(?>(?:(?!\*/|/\*).)*))*).*?\*/|--.*?\r?[\n]", RegexOptions.Singleline);
-            Match commentMatch = regexObj.Match(contents);
-
-            return tokenRegex.Replace(contents, match => ReplaceToken(match, variables, commentMatch));
+            using (var parser = new VariableSubstitutionSqlParser(contents))
```
dazinator: Can this line now be removed? https://github.com/DbUp/DbUp/pull/190/files#diff-d9ae76871876bd2c951280aa2608d84aR19

omnibs: Totally, slipped past me!

src/DbUp/Support/SqlServer/SqlParser.cs
```diff
@@ -0,0 +1,518 @@
...
+            var previousChar = CurrentChar;
+            var buffer = new char[Delimiter.Length - 1];
+
+            // read the rest of the delimiter
+            Read(buffer, 0, Delimiter.Length - 1);
+
+            // Support terminator.
+            var peekChar = PeekChar();
+            char? terminator = null;
+            if (peekChar == ';' || peekChar == '\0')
```
dazinator: consider changing `\0` to `NullChar` constant instead?

omnibs: Done! =]

dazinator:
I think these are all good changes! Nice one :)

omnibs:
Thanks for the review!

Just pushed the suggestions =]

src/DbUp/Engine/Preprocessors/VariableSubstitutionPreprocessor.cs
```diff
@@ -29,47 +33,82 @@ public VariableSubstitutionPreprocessor(IDictionary<string, string> variables)
...
+            {
+                return parser.ReplaceVariables(variables);
+            }
         }
 
-        private static bool IsInComment(Match commentMatch, Match variableMatch)
+        private class VariableSubstitutionSqlParser : SqlParser
```
dazinator: Open question: Do you think there would be any value in making this public? I was just thinking would people potentially want to derive from this class to customer variable substitution behavior with their own preprocessors?

omnibs: Good question. Since the previous VariableSubstitutionPreprocessor wasn't very customizable I didn't think much about this.

I'll expose it with the `'$'` assigned to a virtual property, so a simple use case would be overriding the variable delimiter.

I won't expose the option to change the delimiter on the `VariableSubstitutionPreprocessor` so we don't have to worry about giving any guarantees as to what delimiters can be used without breaking `SqlParser`.

dazinator:
Nice one.

> When we check for delimiters we already peek ahead for the delimiter's length here on IsBeginningOfDelimiter, so I believe that if could never be true.

I think you are correct, good spot sir! I'd be inclined to remove the redundant check. There is plenty of test coverage for the `SqlCommandParser` so we should soon pick up on any regressions caused by such a change :)

omnibs:
Exposed the `VariableSubstitutionSqlParser`. 

The `/Preprocessors` directory might not be the best place for it but IDK, `Support/SqlServer` seemed too far away from `VariableSubstitutionPreprocessor`.

As always, forgot to run the public API tests. Oh, the value of CI, right? =p

dazinator:
All LGTM :+1: great work @omnibs 
Think we are ready for @JakeGinnivan to cast his magic.

JakeGinnivan:
Thanks guys. I am working on a proper automated deploy for Shouldly. Once I get 2.8.0 out the door using that I will switch to getting v4 beta out using a similar technique. Then we can all push the button to release :)

JakeGinnivan:
Likely I will get to this early next week after I finish off shouldly then the DbUp netcore branch :)

PR #194, opened 2016-05-12
FileSystemProvider / New option to include scripts from subdirectories

jesuissur:
The FileSystemProvider now accepts an options argument.  This argument holds the already existing options (encoding and filter) but also adds a new one: IncludeSubdirectories.

Thanks
Phil


jesuissur:
The following test is failing : DbUp.Tests.ApiTests.dbup_has_no_public_api_changes
I guess it's expected because I've exposed a new class (FileSystemProviderOptions) and put obsolete on some constructors of the FileSystemProvider class to use the one with the options argument instead of those overloads.

JakeGinnivan:
Thanks, nice change

Issue #188, opened 2016-04-19
Table '<schema>.schemaversions' doesn't exist


soenneker:
As of dbup and dbup-mysql 3.3.4 I'm getting the exception:

`"Table '<schema>.schemaversions' doesn't exist"`

and the script runner fails.

This occurs when I drop the schema and create a new one. I've downloaded and tried using the latest source from master and still get the same exception. With version 3.3.3 or 3.3.2 it's successful.

simoncuce:
Hope this helps....I had the same issue, but there is a work around. Unless you explicitly set the schema name, the VerfiyTableExistsCommand method will run a query that looks for all schemaversion tables across the whole db, rather than just the schema you are using. As a result, it will fail when creating a new table. The answer is to explicitly add the schema to the MySqlDatabase extension method.Se code below

```
 var upgrader =
                DeployChanges.To
                    .MySqlDatabase(connectionString, **schema)**
                    .WithScriptsFromFileSystem(path)
                    .WithTransactionPerScript()
                    .LogToConsole()
                    .Build();
```

simoncuce:
It would seem that the nuget package 3.3.4 is a little behind the code here. Thus the overloader MySqlDatabase(connectionString, "test1") isn't in the nuget package. Just use the code here

droyad:
PR: #195

droyad:
I'm trying to determine whether this is fixed in `4.0.0`, but I couldn't reproduce it on `3.3.5` or the latest.  I do lack deep MySql knowledge

My Code:
```
var connectionString = "server=127.0.0.1;uid=robert;pwd=taMMedPJ1!;database=test2";

var upgrader =
	DeployChanges.To
		.MySqlDatabase(connectionString)
		.WithScript("Foo", @" DELIMITER $$
CREATE PROCEDURE new_proc4 (IN new_param VARCHAR(256))
BEGIN
INSERT INTO tbl (cell) VALUES (new_param);
END$$
DELIMITER ;")
		.LogToConsole()
		.Build();

var result = upgrader.PerformUpgrade();

if (!result.Successful)
{
	Console.WriteLine(result.Error);
	return;
}
Console.WriteLine("Success!");
```

What I did:
- Create a schema `test`
- Run the above code (setting the right database in the connection string)
- Create a schema `test2`
- Run the above code (setting the right database in the connection string)

It creates the SchemaVersion table in both databases

droyad:
[...]

PR #195, opened 2016-05-13
Fix for #188 Table '<schemaversions>' doesn't exist

simoncuce:
When deploying to a newly created MySql database, but there are existing databases with schemaversions table created (and you don't explicitly define the name of the database), the VerifiyTableExistsCommand falls back on a SQL query that will return a false result. 

This to avoid this, you needs to ensure that the database name is always defined. If you only pass in a connection string, this doesn't happen. I have thus added some functionaility that if no schema is passed in, then look at the connection string. If it finds a schema name in the connection string use it, else default to the schema name as null

### DIFF
--- a/src/DbUp/Support/MySql/MySqlITableJournal.cs
+++ b/src/DbUp/Support/MySql/MySqlITableJournal.cs
@@ -160,7 +160,7 @@ namespace DbUp.Support.MySql
         /// <returns>True if table exists, false otherwise</returns>
         private bool VerifyTableExistsCommand(IDbCommand command, string tableName, string schemaName)
         {
-            command.CommandText = string.IsNullOrEmpty(schema)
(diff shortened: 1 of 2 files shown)


src/DbUp/Support/MySql/MySqlITableJournal.cs
```diff
@@ -160,7 +160,7 @@ private bool DoesTableExist()
         /// <returns>True if table exists, false otherwise</returns>
         private bool VerifyTableExistsCommand(IDbCommand command, string tableName, string schemaName)
         {
-            command.CommandText = string.IsNullOrEmpty(schema)
+            command.CommandText = string.IsNullOrEmpty(schemaName)
```
simoncuce: This is just clean up

PR #202, opened 2016-07-13
Changed default encoding for EmbeddedScriptProvider from ANSI to UTF8

clockwisemusic:
Currently the default encoding for embedded scripts is ANSI. I think this is a bit confusing, as I assumed it was UTF8 - if you try to insert a UTF8 character into a varchar field in your DB it will automatically "convert" it to ANSI - as a result your character is silently corrupted. I suggest this fix.


clockwisemusic:
https://blogs.msdn.microsoft.com/shawnste/2005/03/15/dont-use-encoding-default/

PR #210, opened 2016-08-06
Handle case where executeScalar is null

skjohansen:
I have in a few cases experiences that executeScalar is null if I run DbUp agains a empty database (for the first time, meaning there are no SchemaVersions either).


JakeGinnivan:
Thanks, have merged into v4 branch manually

Issue #213, opened 2016-08-11
Why does WithScriptsAndCodeEmbeddedInAssembly not filter IScript files?


robvanuden:
I am using the following extension method:
public static UpgradeEngineBuilder WithScriptsAndCodeEmbeddedInAssembly(this UpgradeEngineBuilder builder, Assembly assembly, Func<string, bool> filter);

But there is a comment:
filter:
    //     The Sql Script filter (only affects embdeeded scripts, does not filter IScript
    //     files). Don't forget to ignore any non- .SQL files.

Why is that? I can't think up a reason for it, but maybe I forget something. I would be happy to filter the IScript files. Is there a supposed way to do this, or is this simply not implemented yet?

robvanuden:
I fixed my issue for now with a custom ScriptProvider, based on EmbeddedScriptAndCodeProvider.

Differences:
- In the constructor it saves the filter
- In the GetScripts method it applies the filter

JakeGinnivan:
We are putting together v4 right now. If you wanted to submit a PR updating the `EmbeddedScriptAndCodeProvider` that would be great!

robvanuden:
Is it a problem if it's a breaking change relative to v3?
In other words, should I make a different class or can I just change the behavior of the current one?

JakeGinnivan:
It can be a breaking change, it just needs to be documented. We have a few other small breaking changes in there (mainly API though)

JakeGinnivan:
This seems like it is fixing a bug though

robvanuden:
Yeah I consider it a bug too, but maybe some people trust on this bug nowadays ;-)
I will create a PR tonight, with the breaking change (/bugfix) documented in Breaking changes.md

robvanuden:
I created a pull request for this. Please let me know what you think about it.

PR #214, opened 2016-08-11
EmbeddedScriptAndCodeProvider now also applies the filter on code bas…

robvanuden:
…ed scripts #213


RaringCoder:
Any word on this? PR posted a while ago, v4 beta out with the same issue still.

robvanuden:
I don't know, this project seems to be quite much abandoned.
If you encounter this bug, the easiest is to create your own IScriptProvider (copy-paste from this PR) and use it with builder.WithScriptProvider.

RaringCoder:
Yeah, last updated 3 months ago. I would consider it in a "done" state to be honest with a slow move towards .NET Core. What you suggested is pretty much what I did 👍

PR #217, opened 2016-08-26
In Progress -- Basic BatchNumber support

hossambarakat:
This is a work in progress PR to add support for BatchNumber column inside the SchemaVersion

### DIFF
--- a/Breaking changes.md
+++ b/Breaking changes.md
@@ -1,3 +1,4 @@
 # 4.0.0
  - Default Encoding is now UTF rather than Encoding.Default
  - AdHocSqlRunner changed to use Expression<Func<>> rather than Func<>
+ - IJournal includes new property `BatchNumberEnabled` to include BatchNumber column in the journal table
--- a/src/dbup-core/Builder/StandardExtensions.cs
+++ b/src/dbup-core/Builder/StandardExtensions.cs
@@ -470,4 +470,16 @@ public static class StandardExtensions
     {
         return WithScripts(builder, new EmbeddedScriptsProvider(assemblies, filter, encoding));
     }
+
+
+    /// <summary>
+    /// Sets a configuration flag which will cause the engine to include batch number in schema versions.
+    /// </summary>
+    /// <param name="builder">The builder.</param>
+    /// <returns></returns>
+    public static UpgradeEngineBuilder WithBatchNumberEnabled(this UpgradeEngineBuilder builder)
+    {
+        builder.Configure(c => c.Journal.BatchNumberEnabled = true);
+        return builder;
+    }
 }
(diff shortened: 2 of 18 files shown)


JakeGinnivan:
Few things to consider:

 - If someone has enables batch mode on a database which has had scripts run in before. It still needs to work (need tests for this)
 - I think it should either be made compatible (so it just works), or opt-in, not opt out.

docs/more-info/rollback.md
```diff
@@ -0,0 +1,53 @@
+Although roll forward is always the recommended approach. Some systems may require having downgrade scripts.
```
JakeGinnivan: Can we put a section on the downsides of the approach. I think we should not encourage the approach and that people should know the downsides.

Call out things like

 - Make sure you thoroughly test your rollback scripts, they are not always executed so you need to build the confidence they are reliable through testing
 - Ensure you do not drop data, rename columns to _rollback or something to make sure rollback scripts do not incur data loss
 - Rollback scripts add a lot of complexity, make sure that you want to go down this road rather than just taking a backup or snapshot.

Any other downsides you can think of.

docs/more-info/rollback.md
```diff
@@ -0,0 +1,53 @@
...
+
+The verification behaviour could be overrided via the fluent APIs to specify start the verification from version
+
+``` csharp
+    .WithRollbackVerificationFrom("2.0.1")
+```
+
+## Rehearse Downgrade Scripts
+
+It is always recommended to go forward however if you decided to go with the rollback, then it is highly recommended to rehearse the rollback scripts and that could be achieved by performing upgrade then downgrade and then upgrade.
```
JakeGinnivan: It would be good to elaborate on this section. Having automated tests which populate some test data, upgrade then downgrade and make sure no data is lost?

docs/more-info/rollback.md
```diff
@@ -0,0 +1,53 @@
...
+
+``` csharp
+    var upgrader =
+        DeployChanges.To
+            .SqlDatabase(connectionString)
+            .WithScriptsEmbeddedInAssembly(Assembly.GetExecutingAssembly())
+            .WithRollbackEnabled()
+            .Build();
+    
+    Version targetVersion = Version.Parse("1.2.0")
```
JakeGinnivan: I think we need to add a section on versions, maybe even it's own doc page. In the code it is called batches, here it's version. We should also possibly make it arbitrary, for instance a batch could be a sprint?

JakeGinnivan:
Put a few initial comments, good start

JakeGinnivan:
I think we should try to get the Batch support in. @hossambarakat @dennisroche 

It is a good feature which I think would be quite handy for a number of scenarios. Thoughts?

hossambarakat:
@JakeGinnivan I will remove the documentation of the rollback from this PR and will only keep the BatchNumber

dennisroche:
What is needed to get `BatchNumber` in?

hossambarakat:
@dennisroche I have finished the Opt-in for the batch number and fixed the tests, I need to do more testing which I will do tomorrow.

hossambarakat:
@JakeGinnivan Would you kindly review the current PR and provide me with your feedback?

hossambarakat:
I will implement the rest but I wanted to verify the concept in SQL Server and then implement the rest.

PR #221, opened 2016-09-21
Add support for drop database and implementation for SQL Server

dennisroche:
https://github.com/DbUp/DbUp/issues/220

Refactored `EnsureDatabase`, extracting the logic for getting the master connection string.

Issue #220, opened 2016-09-19
Add support for dropping a database


dennisroche:
In our local and development environment, we want to be able to drop the database before running the migration scripts as, for various reasons, we cannot simply roll forward. 

DbUp already has support for ensure database exists. I suggest that we have an equivalent for drop.

I think this has been raised in the past as part of another discussion (can't find the issue atm).

hossambarakat:
I think this could be useful for integration tests as well where you want to drop and recreate the database before the tests

PR #222, opened 2016-09-21
Add support for drop database to v3

dennisroche:
Back ported the changes from https://github.com/DbUp/DbUp/pull/221 to fit against the current `master` branch.

https://github.com/DbUp/DbUp/issues/220


droyad:
I don't think we need to do any more V3 releases

PR #223, opened 2016-09-22
[WIP] Replace IUpgradeLog with LibLog

dennisroche:
`LibLog` is added to the Core and removed implementation of `IUpgradeLog`. 

Need to fix implementations in the providers. Also, a good time to review what is being logged and if the log level is appropriate. 
- [ ] SQL Lite
- [ ] SQL Server
- [ ] Postgresql
- [ ] Firebird
- [ ] SQL CE
- [ ] MySQL

Other tasks
- [ ] Testing
- [ ] Provide samples on setting up logging with Serilog


droyad:
What if instead of replacing the existing logging mechanism, we ship an implementation of IUpgradeLog that uses LibLog and make that the default? This way we would not break existing code.

Would we lose something if we did this? (eg structured logging)

dennisroche:
Let's try that 👍. I forgot that I started this PR as life got busy.

droyad:
Oops, this closed when I deleted the `4.0.0` branch after merging it into `master`

dennisroche:
This can be closed this as this achieves the same https://github.com/DbUp/DbUp/pull/307

Issue #42, opened 2014-01-13
Downgrade functionality


Lars-Kolsaker:
Hi
Great product!

Any plans on implementing functionality for setting the database back to a previous version.

dazinator:
I'd love to see a "target version" concept in DbUp. At the moment there is no concept of database version really, just script names - which you explicitly tell dbup the scripts to run. The database  "version" is simply the name of the last script that was run and recorded in the schema versions table. I'd like instead to tell DbUp the target version number I want to migrate the DB to, and let it decide which scripts are necessary to execute to take the database to that version.

One rudimentary way to outline this idea, imagine creating a subfolder in the projects scripts folder per version of the DB, and putting a single script + rollback script in that folder:

    - 1.0.0
        - addedtable.sql
        - addedtable-rollback.sql
    - 1.0.1
        - addedcolumn.sql
        - addedcolumn-rollback.sql

You could then ask DbUp to migrate the DB to a partiular version:

```
.ToTargetVersion("1.0.1")
```

Which would result in all of the upgrade (or rollback) scripts being run to take the DB to that version.

rickculpepper:
> I'd love to see a "target version" concept in DbUp. At the moment there is no concept of database version really.... I'd like... to tell DbUp the target version number I want to migrate the DB to, and let it decide which scripts are necessary to execute to take the database to that version. 
> --snip---

This would be a useful feature. 

> One... way to... [do this is by] creating a subfolder in the projects scripts folder per version of the DB, and putting a single script + rollback script in that folder....
> 
> --snip--
> You could then ask DbUp to migrate the DB to a partiular version:
> 
> '.ToTargetVersion("1.0.1")'

I like the concept here, though it seems the idea of a single upgrade script and a single rollback script per folder might be limiting. On the other hand, it make the upgrade process very granular, which I like. If you are limiting to one upgrade and one rollback script per folder, then why not mandate the names 'upgrade.sql' and 'rollback.sql'? Hmmm... this idea might not be well-throught-through... ;)

It seems to me that the upgrade side of this can already be done with DbUp using a filter on folders of scripts.

JakeGinnivan:
I don't think we should invent our own separators inside sql files. I think we should try to stick to features already in DbUp to do this.

The way I would do it:

1. Add `.WithRollbackScriptConvention("_rollback")`, we could use optional parameters to fill in the default
1. Add `.PerformDowngrade(target)` on the `UpgradeEngine`
[...]

PR #229, opened 2016-10-25
#42 Downgrade functionality

patrick-tan:
Added support for two forms of rollback

1. Rollback a specific executed script
2. Multiple rollback where you can specify the executed script to roll back to excluding the specified script itself. I.e. if I executed scripts A, B, C and D in that order and I specify to rollback to B then script C and D will be rolled back.

It uses naming convention to find the rollback scripts. E.g. if the rollback suffix is "_rollback" and the executed script name is "test.sql" then the rollback script should be "test_rollback.sql"

Also fixed bug with incorrect column name in SQLiteTableJournal where it had "SchemaVersionID" instead of "Id"


sungam3r:
@patrick-tan Please rebase onto `master` if you are still interested.

mjauernig:
Maybe we should close this. With the extension [DbUp.Downgrade](https://github.com/asimeonov/DbUp.Downgrade) their is a solution. See also [#704](https://github.com/DbUp/DbUp/pull/704)

sungam3r:
> Also fixed bug with incorrect column name in SQLiteTableJournal where it had "SchemaVersionID" instead of "Id"

I would keep it opened for some time at least for this.

Issue #227, opened 2016-10-03
Upgrade Failed: Input string was not in a correct format.


DotNetRockStar:
I keep getting this error but I am having trouble tracking down exactly where/what this is pointing at exactly.  Any ideas why this would be thrown and how I might be able to identify the script that is causing the issue?

Upgrade failed due to an unexpected exception: System.FormatException: Input string was not in a correct format.    at System.Text.StringBuilder.AppendFormatHelper(IFormatProvider provider, String format, ParamsArray args)    at System.String.FormatHelper(IFormatProvider provider, String format, ParamsArray args)    at System.String.Format(IFormatProvider provider, String format, Object[] args)    at System.IO.TextWriter.WriteLine(String format, Object[] arg)    at System.IO.TextWriter.SyncTextWriter.WriteLine(String format, Object[] arg)    at System.Console.WriteLine(String format, Object[] arg)    at DbUp.Engine.Output.ConsoleUpgradeLog.Write(ConsoleColor color, String format, Object[] args) in C:\projects\dbup\src\DbUp\Engine\Output\ConsoleUpgradeLog.cs:line 43    at DbUp.Engine.Output.ConsoleUpgradeLog.WriteError(String format, Object[] args) in C:\projects\dbup\src\DbUp\Engine\Output\ConsoleUpgradeLog.cs:line 27    at DbUp.Support.SqlServer.SqlScriptExecutor.Execute(SqlScript script, IDictionary`2 variables) in C:\projects\dbup\src\DbUp\Support\SqlServer\SqlScriptExecutor.cs:line 135    at DbUp.Engine.UpgradeEngine.PerformUpgrade() in C:\projects\dbup\src\DbUp\Engine\UpgradeEngine.cs:line 71

droyad:
Fixed with the above PR, thanks for submitting.

PR #230, opened 2016-10-29
Fix issues with formatting exceptions.

eKipod:
Fixed issues with exception logging:
1. When a message containing curly braces is raised by the DB.
2. When a SqlException is thrown, the format parameters did not match the actual parameters.

Addresses issues found in #227

### DIFF
--- a/src/DbUp/Support/SqlServer/SqlScriptExecutor.cs
+++ b/src/DbUp/Support/SqlServer/SqlScriptExecutor.cs
@@ -131,21 +131,21 @@ namespace DbUp.Support.SqlServer
             catch (SqlException sqlException)
             {
                 log().WriteInformation("SQL exception has occured in script: '{0}'", script.Name);
-                log().WriteError("Script block number: {0}; Block line {1}; Message: {2}", index, sqlException.LineNumber, sqlException.Procedure, sqlException.Number, sqlException.Message);
-                log().WriteError(sqlException.ToString());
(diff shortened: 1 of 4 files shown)


src/DbUp/Support/SqlServer/SqlScriptExecutor.cs
```diff
@@ -131,21 +131,21 @@ public virtual void Execute(SqlScript script, IDictionary<string, string> variab
...
                 log().WriteError("Script block number: {0}; Error code {1}; Message: {2}", index, sqlException.ErrorCode, sqlException.Message);
-                log().WriteError(sqlException.ToString());
+                log().WriteError("{0}", sqlException.ToString());
                 throw;
             }
             catch (Exception ex)
             {
                 log().WriteInformation("Exception has occured in script: '{0}'", script.Name);
-                log().WriteError(ex.ToString());
+                log().WriteError("{0}", ex.ToString());
```
robvanuden: What is exactly the point of adding all those "{0}",  ?

eKipod: The first argument to ``WriteError()`` is a format. ``ex.ToString()`` is an arbitrary string, from an external source (i.e. the DB). If it contains a format string itself (e.g. curly braces), the result is that the ``WriteError`` operation will fail (with an ``InvalidFormat`` exception). Adding these "{0}", makes sure that the format part is a indeed a format, and the result is a valid string, eliminating the need to "escape" or otherwise sanitize  the input (or add a separate interface that takes no parameters). 
I added a unit test for an example (and validation).

ffMathy:
I have this issue as well. When will this be merged in?

PR #231, opened 2016-11-03
Additional FileSystemScriptProviderOptions

rintje:
Hi Jake,

A while ago we discussed separating sorting functionality in #187. Good call, making me wonder if a separate VersionFolderScriptProvider is even necessary. 
Turns out, with two extra filesystem provider options and the IncludeSubDirectories option, one can realize the VersionFolderScriptProvider functionality with just the fileystem script provider:

1. Comparer: supply a custom class implementing IComparable<string> to sort the scripts in any way you like, for example by Version or by SemVer.

2. Scriptnamer: supply a scriptnamer delegate to define a custom name for the scripts, to be more granular when distinguishing between identically named files in different folders when using option IncludeSubDirectories. 
Note: Default naming is still filename to be backwards compatible.

Let me know what you think.
Regards


droyad:
@rintje  First of all, sorry about the delay in attending to this.

I think item 1 is covered in #314, you can supply a custom comparer

Once #312 is merged, I'll pull in the naming code

SimonRupar:
Hello, 
We would like a possibility to provide some sort of "Script namer". Because current functionality provided in db4-beta0004 with recursive paths works perfect for us, but we are still using filenames only (without attached path). So at this point, we are forced to generate a collection of SqlScript ourself without using native providers.
Conflicting part:
`
#SqlScript.cs
            var filename = fullPath
                .Substring(fullBasePath.Length)
                .Replace(Path.DirectorySeparatorChar, '.')
                .Replace(Path.AltDirectorySeparatorChar, '.')
                .Trim('.');
`

Best Regards, 
Simon Rupar

sungam3r:
@rinkeb  Please rebase onto `master` if you are still interested.

PR #233, opened 2016-12-09
Adding support for Azure SQL Database Creation Edition

JustinMason:
Adding support to provide the Azure Edition to the EnsureDatabase extension.

Azure has some extra parameters.  We usually run a local SQL Server Instance on build and development machines.  I leaned toward including Azure in the existing SQLServer extension rather than have a separate AzureSqlServerExtension.  I could see a case for a separate one that enables more Azure specific abilities but for simple database creation I liked the simplicity.


droyad:
@JustinMason Thank you, sorry about the delay

PR #241, opened 2017-02-17
Add event after script is executed

LmPierson:
I wanted the ability to do some post-processing of each script after the script is executed within my local application.  This code change simply fires an event each time a script is executed.  It contains information about the script as well as the connection manager.  I use the connection manager to execute within the connection in my local app.


LmPierson:
Not sure from the log what failed.  Please let me know what you need me to do.

AdrianJSClark:
DbUp has tests that check to make sure you do not accidentally change the public API surface. Your build failed because adding your event changes the API and you did not update the test to expect that.

Here is the detail from [the build log](https://ci.appveyor.com/project/DbUp/dbup/build/3.3.6-PullRequest.241%2b20.build.252):
```
Errors and Failures:

1) Test Failure : DbUp.Tests.ApiTests.dbup_has_no_public_api_changes
     Expected string length 30470 but was 31366. Strings differ at index 46.
  Expected: "...m.CLSCompliantAttribute(true)]\n[assembly: System.Runtime.I..."
  But was:  "...m.CLSCompliantAttribute(true)]\r\n[assembly: System.Runtime...."
  ---------------------------------------------^
```

If you run the tests locally you should have the ability to update the file that defines the accepted API verified by that approval test with the new value.

droyad:
Superceeded by #415

Issue #235, opened 2016-12-26
It should also log number of rows affected !


juanonsoftware:
In SqlScriptExecutor method Execute I see we has a call: command.ExecuteNonQuery();

-> Should we log it ? I see a case that this number is also useful ! Please consider to take this into log !
-> Should we log result using a different (new) method instead of WriteInformation ? This way the custom Log can handle this more flexible.

AdrianJSClark:
If you configure your upgrader with `.LogScriptOutput()` the `SqlScriptExecutor` will use `ExecuteQuery()` and include the output of any result sets in the logs. Obviously this isn't exactly what you were looking for, however ending your script with some SQL similar to `SELECT @@ROWCOUNT as [Rows Affected by Changes]` would cause this value to be written using this method.

droyad:
Fixed by #416

PR #244, opened 2017-03-11
#235 Added logging for rows affected.

osfancy:


droyad:
Resolves #235

sungam3r:
Outdated, #416

PR #247, opened 2017-03-27
Update README.md

cskwrd:
NuGet download counts are not currently supported on Shields.io. This is due to the fact that the NuGet Gallery API doesn't currently provide this information.

See https://github.com/badges/shields/issues/655


droyad:
The download counts seem to work now.

Issue #248, opened 2017-03-30
Firebird constantly trying to create SchemaVersions table


Brett-Gladman:
When using DeployChanges with the Firebird database, I've found that the engine attempts to create the SchemaVersions table after each script.
It appears that it always fails to find SchemaVersions as an existing table.

I tracked down the issue to the FirebirdTableJournal within the VerifyTableExistsCommand method.

It is currently using :

"select 1 from RDB$RELATIONS where RDB$SYSTEM_FLAG = 0 and RDB$RELATION_NAME = '{0}'" 

to determine the existence of a table.

Firebird 3 (and I suspect 2.5) force RDB$RELATIONS.RDB$RELATION_NAME to uppercase, so a lower case tableName passed through always fails.

I found this change to work:

"SELECT DISTINCT 1 FROM RDB$RELATIONS WHERE RDB$SYSTEM_FLAG = 0 AND UPPER(RDB$RELATION_NAME) = UPPER('{0}')"

Timothyoverton:
I found the same fix, has a pull request been made/merged for this change?

Timothyoverton:
Here is my pull request for the fix: https://github.com/DbUp/DbUp/pull/250

Brett-Gladman:
Tim,
  I was evaluating DbUp as a possible addition to an existing project, so wasn't really looking to fix anything, just reporting an encountered issue. Rather than change the code above, I found it easier just to call JournalToFirebirdTable("SCHEMAVERSION") off DeployChanges, which passed through the required uppercase version every time.

Timothyoverton:
Hey Brett, thanks for the alternate solution - I'll try the same, cheers!

Issue #243, opened 2017-03-08
FirebirdSampleApplication not working - unsuccessful metadata update


Timothyoverton:
Hi Guys,

I'm just trying to get the sample firebird application running.

I have left the nugets as per sample:
`<package id="dbup" version="3.3.5" targetFramework="net452" />
  <package id="dbup-firebird" version="3.3.5" targetFramework="net452" />
  <package id="FirebirdSql.Data.FirebirdClient" version="4.5.2.0" targetFramework="net452" />`

I'm running Firebird 3, 64bit, 3.0.0.32483.

I have a populated db and I'm trying to run 1 script:
`CREATE TABLE Testing1 (ID int NOT NULL primary key);`

The schema table is being created and the Testing1 table is being created, then it fails on:
> at DbUp.Engine.UpgradeEngine.PerformUpgrade() in C:\projects\dbup\src\DbUp\Engine\UpgradeEngine.cs:line 73

> No message for error code 336397285 found.
>No message for error code 336068862 found. ---> unsuccessful metadata update
>No message for error code 336397285 found.
>No message for error code 336068862 found.

I've tried commenting out some settings, I always get the Testing1 table created, with this line in place:
`.WithTransactionPerScript()`
I don't get the schema table created, if I comment out this line, I do...
including `.WithoutTransaction()` has no effect.

Any ideas before I go hunting into the DBup code? What version of Firebird was this Sample working for?

Timothyoverton:
Anyone?

Timothyoverton:
This issue can also be resolved with my fix here: https://github.com/DbUp/DbUp/pull/250

hhindriks:
In intend to fix any issues with the FirebirdSampleApplication and submit a PR for it soon. Currently I need to comment out both 
` 
.WithTransactionPerScript() and 
.WithTransaction()
`
for the application to run through.

Jeern:
@mjauernig  and @sungam3r  This was fixed by my PR https://github.com/DbUp/DbUp/pull/692

I have tested that the SampleApplication works. However

1. WithTransaction is no longer allowed (AllowedTransactionMode)
2. WithTransactionPerScript did not work before my fix, but works now.
3. I need to upgrade the FirebirdSql.Data.FirebirdClient to the latest to make it work for FIrebird versions newer than 2.5 (because the old Nuget package does not support Wire encryption)

I will provide af PR for nr. 3 - after that this Issue can be closed

Jeern:
[...]

PR #250, opened 2017-04-29
Fix to VerifyTableExistsCommand method, lower case tablename fails verify

Timothyoverton:


Timothyoverton:
Hi All, my PR is failing because of line endings, not sure how to resolve?

yufan-at-arcadlon:
You are using Windows LE while the test is expecting UNIX LE. Git can be configured to automatically translate the LE and make sure only UNIX LE would be committed. You should set `autocrlf` to `true` like this:

    git config --global core.autocrlf true

https://help.github.com/articles/dealing-with-line-endings/

sungam3r:
@Timothyoverton Please rebase onto mastwr if you are still interested.

mjauernig:
Maybe it is already fixed. See #243

Jeern:
@sungam3r  and @mjauernig  Correct. It is fixed (see my comment on #248 )

PR #253, opened 2017-05-17
Added support for Azure SQL Data Warehouse.

dameeks:


droyad:
Happy to accept another PR for this now that 4.0 is out. See the recent Oracle one: https://github.com/DbUp/DbUp/pull/326

dameeks:
I updated the Azure SQL Data Warehouse implementation to play nice with 4.0.

annhwallinger:
Is there a timeline to include this? With SSDT in preview support of Azure SQL Data Warehouse, this would be a GREAT fit!

madelson:
Any updates on this?

jbradysmith:
for anyone following this request, we were able to get DBUp working with the Azure SQL DW by createing our own schema table. The connectionm string is just passing the DB UN and PWD. There may be unkiown consequences down the road but we will see how this goes

CREATE TABLE [dbo].[SchemaVersions]
(
	[Id] [int] IDENTITY(1,1) NOT NULL,
	[ScriptName] [nvarchar](255) NOT NULL,
	[Applied] [datetime] NOT NULL
)
WITH
(
	DISTRIBUTION = HASH([ScriptName]),
	CLUSTERED COLUMNSTORE INDEX
)
GO

lsilva-tyl:
Attempting to run against a Synapse SQL pool. The schema gets created but when creating a table using default values I get 
`Script block number: -1; Message: Enforced unique constraints are not supported. To create an unenforced unique constraint you must include the NOT ENFORCED syntax as part of your statement.`

sungam3r:
@dameeks  Please rebase onto `master` if you are still interested.

Issue #257, opened 2017-06-19
More information in DatabaseUpgradeResult....


natiki:
Hi,

In my particular case we run a number of PerformUpgrade() calls against a few databases when our application starts up. 

While we can work out from the scripts collection what succeeded and what didn't it would be great if we could supply some form of logical name to the UpgradeEngineBuilder and then get that back in the result. So something like:

```c#
UpgradeEngineBuilder upgradeBuilder =
                DeployChanges.To
                             .SqlDatabase(_ConnectionString)
                             .MigrationLogicalName("MyLogicalName");
```
```c#
if (upgradeEngine.PerformUpgrade().MigrationLogicalName == "MyLogicalName")
{

}
```

natiki:
Bump!

AdrianJSClark:
Hi @natiki,

Just wondering if you could expand on this idea a little. Is there something that this tag can give you which a separate instance of the upgrader cannot?

Where else would you expect to see the `MigrationLogicalName` value appear? In the Journal table, for example?

natiki:
Hi @AdrianJSClark,

Apologies for the heaps of code, but hopefully this illustrates things a bit better.

We use a simple wrapper DatabaseMigrationsManager to migrate a number of databases all in one go using our thin wrapper DatabaseMigrationsRunner around DbUp.

```c#
    /// <summary>
    ///     Class to manage migrating one or more physical databases. It uses the DatabaseMigrationsRunner to do the heavy lifting.
    /// </summary>
    public class DatabaseMigrationsManager
    {
        public delegate void OnErrorMessageHandler(object aSender, MessageErrorEventArgs aArgs);

        public delegate void OnInformationMessageHandler(object aSender, MessageInformationEventArgs aArgs);

        public delegate void OnWarningMessageHandler(object aSender, MessageWarningEventArgs aArgs);

        /// <summary>
        ///     Based on the host name supplied we make a determination of what database kind we are interacting with.
        ///     The determination is very restrictive in that if we do not find a match then we raise an exception.
        /// </summary>
        /// <param name="aHostName">Hostname to use to determine the db environment.</param>
        /// <returns>Returns the database environment if found.</returns>
        public static DbEnvironmentKindEnum DetermineDbEnvironmentFromHostName(string aHostName)
        {
            string hostName = aHostName.Trim().ToLower();

            if (hostName.StartsWith("www.") || hostName.StartsWith("m.")
[...]

PR #261, opened 2017-07-20
#257 - added UpgradeEngineBuilder.MigrationLogicalName property, and …

bcwiniger:
…StandardExtensions.SetMigrationLogicalName extension to set the property


bcwiniger:
Can someone explain how to contribute to this project if the public API diff test fails? I am new to this and would like to know what I am doing wrong.

dazinator:
The public API test is an approval test - it compares the generated file describing the current API against the previous one which is the "golden master" definition. The test is there so you can more easily review if there are any breaking changes introduced. If there are breaking changes then it means we should bump the major version number for the release. I haven't reviewed this PR but if you havent introduced any breaking api changes then there is nothing to worry about. To get the test to pass, copy the contents of the generated file in place of the contents of the current "golden master" file so that the next time the test runs it will compare the API surface to itself = pass.

bcwiniger:
Thanks for the explanation. Since there are no breaking changes, just an addition, should I fix the test and resubmit?

dazinator:
@bcwiniger - yep!

droyad:
@bcwiniger Appologies for us taking so long to look at this. What is the use case for this property?

droyad:
I think this is related to #257

sungam3r:
@bcwiniger Please rebase onto mastwr if you are still interested.

Issue #249, opened 2017-04-11
Applied time in DB is not UTC


mathijsco:
When dbup is run, the SchemaVersions table is created automatically.
Also each version of the scripts that has been run, are added to that table.

However the time that is stored in the table is not UTC. 
The code in SqlTableJournal uses "DateTime.Now", but should use "DateTime.UtcNow".

Is there a reason why it uses the local time and not storing the time zone? Or is this a bug?

droyad:
No reason I am aware of, but I hesitate to change the default as it would make the data in existing `SchemaVersion` tables inconsistent. Perhaps this could be a configuration option?

Even better would be making the column a `datetimeoffset`, but I'm unsure how that translates to the other database platforms.

AdrianJSClark:
The original reason for simply using local time was likely just simplicity. I agree with @droyad that it would be unwise to simply change the default. A configuration option is really the only way at this stage.

I had a quick idea that if the column type was changed then that could be used to detect the difference between the old and new schemas which could then inform the default. That way new projects with new `datetimeoffset` columns get UTC by default, but existing projects with `datetime` columns get local. This path is probably narrow, rocky, and liable to cause twisted ankles though so we should give it heaps more thought than the few minutes I have done.

Perhaps configuration option now, with something more advanced and/or a changed default in a future version?

droyad:
The whole things could be contained inside an implementation of `IJournal`, either an option on the existing implementations, or subclass of say `SqlTableJournal`

AdrianJSClark:
Argh! Of course. Silly me, sorry. That's pretty easy then. Just `.JournalTo(new UtcSqlTableJournal())` 👍

droyad:
Do this as part of the advanced Journaling feature

davidspiess:
I still think UTC always should be the default.
If the project uses Semver for versioning, this change could go alongside with the 5.0.0 release and count as a breaking change.

PR #267, opened 2017-08-09
Use utc dates - see #249

SaphuA:
#249


droyad:
@SaphuA Thank you for the PR, however this would make existing tables inconsistent. However, this could be an opt-in feature and perhaps use `datetimeoffset`. See the discussion on #249.

PR #272, opened 2017-09-11
Correct sample code of document.

kazuhito-m:
Replaced it with the code actually executed.


AdrianJSClark:
Thanks for submitting this!

Issue #271, opened 2017-09-08
If an unexpected exception occurs during checking for Journaling table then DbUp re-runs all scripts and potentially destroys lots of data 


markoueis:
Luckily we caught this in a test environment. A SqlException (don't know what type it was but I assume a timeout because the SQL Server database was being hammered at the time) was thrown when DBUp was checking for the existence of the Journaling table. As a result, it assumed the table did not exist and that the database was at version 0. It then proceeded to run all historical migration scripts which of course screwed things up pretty badly.

In looking at the code to diagnose what went wrong we can see that if any SqlException or DbException is thrown during checking for the table (timeouts are including in SqlExceptions), then DbUp proceeds as though table doesn't exist. See code below
 
```
 public string[] GetExecutedScripts() // this is the method called to see what previous scripts ran
        {
            log().WriteInformation("Fetching list of already executed scripts.");
            var exists = DoesTableExist(); //THIS IS THE PROBLEM METHOD. SEE IMPL BELOW
            if (!exists)
            {
                log().WriteInformation(string.Format("The {0} table could not be found. The database is assumed to be at version 0.", CreateTableName(schema, table)));
                return new string[0];
            }
```

```
private bool DoesTableExist()
{
   return connectionManager().ExecuteCommandsWithManagedConnection(dbCommandFactory =>
   {
      try
      {
         using (var command = dbCommandFactory())
         {
            return VerifyTableExistsCommand(command, table, schema); //this just calls SQL
          }
       }
       catch (SqlException) //AS YOU CAN SEE, ANY SQL EXCEPTION including TIMEOUT would be caught here
       {
          return false;
       }
       catch (DbException)
       {
          return false;
       }
            });
        }
```

This seems like a critical issue to me. Is there a way around this?

PR #276, opened 2017-09-13
Removed exception handlers from the `DoesTableExist` method

droyad:
Removed exception handlers from the `DoesTableExist` method and allowed exceptions to propogate. Fixes https://github.com/DbUp/DbUp/issues/271


markoueis:
Hey guys. Thanks for this work. I'm a little new the dbup community but when will this be merged in and in which release will it make it in?

Thanks

PR #291, opened 2017-11-02
Revert "Removed exception handlers from the `DoesTableExist` method"

droyad:
Reverts DbUp/DbUp#276

I merged this, but then tried updating this for 4.0.0 and realised that this area of code has changed significantly and this fix is not needed. I'm not intending to create a new release for 3.x, so reverting this means the merge to 4.0 will be easier


markoueis:
Does this mean the 4.0 final release will have this fix?

droyad:
As far as I can see `4.0` doesn't have this problem as this area was redone.

Issue #198, opened 2016-05-25
No failure if unexecuted script is less than executed scripts


magitech:
Scripts will run in lexicographic order but do not respect if a later script has already been run. For example, if the database has been upgraded with Script1.sql then Script0.sql is added DbUp will attempt to run Script0.sql on the next upgrade. My expectation would be such script would not be executed and warnings/errors would be logged. As a nice to have create an override to disable skipping to allow for backwards compatibility.

JakeGinnivan:
Completely agree, this would be a great addition.

Would you be able to submit a PR for v4? (release/4.0.0 branch)

robvanuden:
I can pick this up if you're ok with that!

droyad:
I disagree with this, say two people are working on a branch, the first commits `Script0001 - Foo` the other `Script0001 - Bar`. If the former is merged to `master` before the latter, the latter script will never run. The first one could have gone to production. The second person would need to check to see if that is the case and then rename the script.

This scenario happens all the time for us and the current situation all works pretty seamlessly as the scripts usually modify different parts of the system.

If this is something people want, it should be opt-in

dazinator:
I agree with @droyad, it also means the schemaversion table would end up looking different on different environments, for example some would have executed script 0 and others would not have. This kind of violates the principles behind DbUp in my opinion, and therefore having it an "opt in" feature would be better.

wozzo:
It is opt in. The additional config val needs setting or it does nothing.

droyad:
@wozzo Awesome

mjauernig:
Seems to be done already. If still needed, please reopen.

PR #302, opened 2018-01-28
GH-198 Add EnforceScriptOrder

wozzo:
Ensures that scripts are executed in correct order.
If there are previously executed scripts, ensures that no other scripts are scheduled to run ahead of them, in the current order as provided by the script provider.
Saw on issue #198 that it was ok to pr into this branch a year and a half ago. Assuming that's still the case.

### DIFF
--- a/src/dbup-core/Builder/StandardExtensions.cs
+++ b/src/dbup-core/Builder/StandardExtensions.cs
@@ -368,6 +368,17 @@ public static class StandardExtensions
     }
 
     /// <summary>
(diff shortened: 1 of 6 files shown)


droyad:
@wozzo I've commented on the #198

droyad:
I'm closing this for now, as I think an alternate approach would make it easier. I've submitted #311 which allows the addition of a script filter. That script filter can be used to implement this feature, what do you think? I'll have a crack at doing so in the next week or two, unless you get to it before me.

PR #308, opened 2018-02-07
Ordered all scripts by name. Sped up the comparison of executed scripts.

### LINKED ISSUE #180: Script ordering not using natural sorting
After integrating DbUp into our deployment process, we just noticed a case where scripts have been run out of order. E.g. our SchemaVersions table shows:
- Our.Namespace.Scripts.100. Some interesting script name here.sql
- Our.Namespace.Scripts.89. This script should appear after 100.sql

Looking at the source, the [EmbeddedScriptsProvider](https://github.com/DbUp/DbUp/blob/master/src/DbUp/ScriptProviders/EmbeddedScriptsProvider.cs) is using `OrderBy(sqlScript => sqlScript.Name)` which won't apply natural sorting.

The other providers don't appear to do any sorting, so probably rely on the ordering provided by the file system. Is there going to be a conflict of licensing if I add support for natural sorting using Dave Koelle's implementation, [AlphanumComparator.cs](http://www.davekoelle.com/files/AlphanumComparator.cs)?

droyad:
Related to #298 
Resolves #180
I'm pretty sure the new default behaviour of running all scripts in order would be the expected behaviour instead of running each provider at a time.

### DIFF
--- a/src/dbup-core/Engine/IScriptSorter.cs
+++ b/src/dbup-core/Engine/IScriptSorter.cs
@@ -1,0 +1,9 @@
+﻿using System.Collections.Generic;
+
+namespace DbUp.Engine
+{
+    public interface IScriptSorter
+    {
+        IEnumerable<SqlScript> Sort(IEnumerable<SqlScript> filtered);
+    }
+}
--- a/src/dbup-core/Builder/StandardExtensions.cs
+++ b/src/dbup-core/Builder/StandardExtensions.cs
@@ -330,7 +330,21 @@ public static class StandardExtensions
     {
         return WithScripts(builder, new EmbeddedScriptAndCodeProvider(assembly, filter));
     }
-
+    
+    /// <summary>
+    /// Adds a sorter that sorts the scripts before filtering and execution
+    /// </summary>
+    /// <param name="builder">The builder.</param>
+    /// <param name="sorter">The sorter.</param>
+    /// <returns>
+    /// The same builder
+    /// </returns>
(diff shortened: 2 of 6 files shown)


src/dbup-core/Engine/IScriptSorter.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Collections.Generic;
+
+namespace DbUp.Engine
+{
+    public interface IScriptSorter
+    {
+        IEnumerable<SqlScript> Sort(IEnumerable<SqlScript> filtered);
```
AdrianJSClark: Should this return `IOrderedEnumerable<SqlScript>` ? Not sure what difference that makes, but sorting definitely puts the scripts in an order.

AdrianJSClark:
@droyad You say this "Resolves" #180 however I'm not sure that is true. From what I've seen here, this would allow #180 to be resolved in the future by supplying a sorter which doesn't just perform straight lexicographical sort. It doesn't fix the specific issue itself, right?

droyad:
@AdrianJSClark Keeping me on my toes. I think this resolves in the way the discussion on that issue ended up, that we should add an extension point.

src/dbup-core/Engine/IScriptSorter.cs
droyad: This would mean though that you couldn't so some sort of sorting and return a list/array. You'd have to convert that to an `IOrderedEnumerable` somehow.

gautamksr:
Can you publish this feature , this is very helpful.

Issue #309, opened 2018-02-07
The scripts created by FileSystemScriptProvider should include subfolder names


droyad:
We should put the subfolder names into the script names created by FileSystemScriptProvider as then we can sort them properly later.

Subfolder support was added in 4.0.0-beta in #194.

PR #312, opened 2018-02-07
Added the subfolder name to script names returned from FileSystemScriptProvider

droyad:
Fixes https://github.com/DbUp/DbUp/issues/309. The subfolder is new in 4.0, so not considering this a breaking change.

### DIFF
--- a/src/dbup-core/Engine/SqlScript.cs
+++ b/src/dbup-core/Engine/SqlScript.cs
@@ -67,10 +67,17 @@ namespace DbUp.Engine
         /// <returns></returns>
         public static SqlScript FromFile(string basePath, string path, Encoding encoding)
         {
-            var filename = Path.GetFullPath(path)
-                .Substring(Path.GetFullPath(basePath).Length)
-                .Trim(Path.DirectorySeparatorChar)
-                .Replace(Path.DirectorySeparatorChar, '.');
+            var fullPath = Path.GetFullPath(path);
+            var fullBasePath = Path.GetFullPath(basePath);
+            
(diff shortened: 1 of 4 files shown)


src/dbup-core/Engine/SqlScript.cs
```diff
@@ -56,12 +56,24 @@ public static SqlScript FromFile(string path)
...
+        /// </summary>
+        /// <param name="basePath">Root path that was searched</param>
+        /// <param name="path">Path to the file</param>
+        /// <param name="encoding"></param>
+        /// <returns></returns>
+        public static SqlScript FromFile(string basePath, string path, Encoding encoding)
         {
+            var filename = Path.GetFullPath(path)
+                .Substring(Path.GetFullPath(basePath).Length)
+                .Trim(Path.DirectorySeparatorChar)
```
AdrianJSClark: Was the `Trim(...)` call necessary here? Should you also be trimming `Path.AltDirectorySeparatorChar`?

src/dbup-core/Engine/SqlScript.cs
```diff
@@ -56,12 +56,24 @@ public static SqlScript FromFile(string path)
...
+                .Replace(Path.DirectorySeparatorChar, '.');
```
AdrianJSClark: Ditto for the replace. What if I've used `Path.AltDirectorySeparatorChar` to separate parts of my path?

src/dbup-core/Engine/SqlScript.cs
droyad: `basePath` does not end with a `/`, so the substringed path ends up with a leading `/`

Issue #334, opened 2018-05-11
Oracle DB Error Attempting to Create or Replace Package


AdrianJSClark:
_Note: Originally posted by @steflege here: https://github.com/DbUp/DbUp/issues/224#issuecomment-388166209_

> Great solution thanks.
> 
> A simple query in an sql script file run well.
> 
> Just have some problems with my create or replace package scripts(utf8 with bom). 
> 
> It always returns :
> ----
> Spec	1	71	PLS-00103: Symbole "end-of-file" rencontré à la place d'un des symboles suivants :
>    . @ % ; is authid as cluster order using external character
>    deterministic parallel_enable pipelined aggregate
>    result_cache accessible
> ---
>  There is the script:
> CREATE OR REPLACE PACKAGE PCK1 AS
> 
> FUNCTION Test(p_param IN VARCHAR2) RETURN VARCHAR2;
> 
> END;
> /
> 
> When I look at the generated Package (with errors) the source looks like :
> CREATE OR REPLACE PACKAGE PCK1 AS  FUNCTION Test(p_param IN VARCHAR2) RETURN VARCHAR2
> /
>  
> It seems to stop where the first ';' character was found in the original script.
> 
> Is it normal? 
> 

@treesrc replied: https://github.com/DbUp/DbUp/issues/224#issuecomment-388279459

> ; (semicolon) is used as a delimiter so I think it is not possible. You can use also c# code for migration. In SampleApplication project there is Script0005 - ComplexUpdate.cs file. In your c# class, you can execute whatever you want.

@steflege responded to this: https://github.com/DbUp/DbUp/issues/224#issuecomment-388347425

> Thank you for your fast answer.
> 
> Unfortunatly, it would be hardly feasible to transform our oracle packages (from 100 to 800 lines) in c# class. Almost all of our scripts are packages replacement(+/- 200 packages per year) or Oracle objects upgrading scripts containing many semicolons.
> 
> There most be an other solution.

ttufekci:
I will change the delimiter for oracle to slash character **/** , if it does not break any test and make a pull request. So we can create trigger, procedure, package etc.

ttufekci:
@AdrianJSClark, @steflege I have made a pull request #335 to use slash character as default delimiter.

sungam3r:
Covered by #509.

Issue #224, opened 2016-09-27
Support for Oracle Database


vengi83644:

pizycki:
there is this PR https://github.com/DbUp/DbUp/pull/228

thebrianlopez:
Any updates on this? Would be awesome to get this feature into this wonderful tool ! 

Thanks for all of your efforts on this!

droyad:
Closed via #326 (thanks @treesrc)

droyad:
I've pushed up the oracle package as a preview: https://www.nuget.org/packages/dbup-oracle

Please let me know if you have success with it

wazbekker:
I tried testing the oracle package connecting to an Oracle 11g database and I am getting the following error:

Declaration referenced in a method implementation cannot be a final method.  Type: 'DbUp.Oracle.OracleTableJournal'.  Assembly: 'dbup-oracle, Version=4.0.1.0, Culture=neutral, PublicKeyToken=null'.

StackTrace:
   at DbUp.Oracle.OracleExtensions.<>c__DisplayClass4_0.<OracleDatabase>b__2(UpgradeConfiguration c)
   at DbUp.Builder.UpgradeEngineBuilder.Build()

droyad:
@wazbekker, did you reference the same version of the dbup-core package? It might be picking up the 4.0.1 release which doesnt have the right changes

wazbekker:
@droyad, sorry my bad.  I changed my dbup-core version to 4.0.1-oraclesupport0001 and my tests were successful.

treesrc:
@droyad, I have tested the nuget package using visual studio 2015, framework 4.5 app and using visual studio 2017 core app. So far no issues.

droyad:
Awsome, thanks @wazbekker @treesrc , I've merged and it should be out as 4.1.0 shortly

steflege:
Great solution thanks.

A simple query in an sql script file run well.

Just have some problems with my create or replace package scripts(utf8 with bom). 

It always returns :
----
Spec	1	71	PLS-00103: Symbole "end-of-file" rencontré à la place d'un des symboles suivants :
   . @ % ; is authid as cluster order using external character
   deterministic parallel_enable pipelined aggregate
   result_cache accessible
---
 There is the script:
CREATE OR REPLACE PACKAGE PCK1 AS

FUNCTION Test(p_param IN VARCHAR2) RETURN VARCHAR2;

END;
/

[...]

PR #335, opened 2018-05-13
Change delimiter from semicolon to slash to support package, trigger, procedure creation

ttufekci:
Related with #334 and #224, I will change delimiter character from semicolon (;) to slash (/). Slash character is more appropriate, because of pl/sql block statements.

Semicolon works for sample sql statements like insert, update, delete, create etc. 

But for creating trigger, procedure, package etc. we have to use multiple statements with semicolon delimiter, and slash for the last statement.
Slash tells where the entire procedure ends and execute it as a unit rather than executing the individual statements inside.

Example script (script001.sql):

```
CREATE TABLE products 
(
  product_id number(10) NOT NULL,
  product_name varchar2(50) NOT NULL,
  product_category varchar2(1) NOT NULL,
  product_date DATE,
  CONSTRAINT products_pk PRIMARY KEY (product_id)
)
/
create sequence product_seq
/
insert into PRODUCTS (PRODUCT_ID, PRODUCT_NAME, PRODUCT_CATEGORY, PRODUCT_DATE) values (PRODUCT_SEQ.nextval, 'test', 'T', Current_date)
/
insert into PRODUCTS (PRODUCT_ID, PRODUCT_NAME, PRODUCT_CATEGORY, PRODUCT_DATE) values (PRODUCT_SEQ.nextval, 'test', 'Z', Current_date)
/

CREATE OR REPLACE TRIGGER orders_before_insert
BEFORE INSERT
   ON orders
   FOR EACH ROW

DECLARE
   v_username varchar2(10);

BEGIN

   -- Find username of person performing INSERT into table
   SELECT user INTO v_username
   FROM dual;

   -- Update create_date field to current system date
   :new.create_date := sysdate;

   -- Update created_by field to the username of the person performing the INSERT
   :new.created_by := v_username;

END;
/

```

### DIFF
--- a/src/dbup-oracle/OracleCommandReader.cs
+++ b/src/dbup-oracle/OracleCommandReader.cs
@@ -13,7 +13,7 @@ namespace DbUp.Oracle
         /// <summary>
         /// Creates an instance of MySqlCommandReader
         /// </summary>
-        public OracleCommandReader(string sqlText) : base(sqlText, ";", delimiterRequiresWhitespace: false)
+        public OracleCommandReader(string sqlText) : base(sqlText, "/", delimiterRequiresWhitespace: false)
         {
         }
 
@@ -25,8 +25,8 @@ namespace DbUp.Oracle
             get
             {
                 string statement;
-                return TryPeek(DelimiterKeyword.Length, out statement) &&
-                       string.Equals(DelimiterKeyword, statement, StringComparison.OrdinalIgnoreCase);
+                return TryPeek(DelimiterKeyword.Length - 1, out statement) &&
+                       string.Equals(DelimiterKeyword, CurrentChar + statement, StringComparison.OrdinalIgnoreCase);
             }
         }
 
--- a/src/dbup-tests/Support/Oracle/OracleConnectionManagerTests.cs
+++ b/src/dbup-tests/Support/Oracle/OracleConnectionManagerTests.cs
@@ -25,9 +25,9 @@ namespace DbUp.Tests.Support.Oracle
         [Fact]
         public void CanParseMultilineScript()
         {
-            var multiCommand = "create table FOO (myid INT NOT NULL);";
+            var multiCommand = "create table FOO (myid INT NOT NULL)/";
             multiCommand += Environment.NewLine;
-            multiCommand += "create table BAR (myid INT NOT NULL);";
+            multiCommand += "create table BAR (myid INT NOT NULL)/";
 
             var connectionManager = new OracleConnectionManager("connectionstring");
             var result = connectionManager.SplitScriptIntoCommands(multiCommand);


AdrianJSClark:
Thanks for putting this PR up, @ttufekci. It looks like good work.

I've done some reading up on the Oracle SQL format and that haphazard Googling-with-Bing suggests to me that the "/" character is the better analog for the "GO" command that is used in SQL Server scripts than the ";" character. However I personally don't have the experience to know if making this change is the correct course of action, or if it will cause existing DbUp users, who are relying on the old behaviour, to be broken.

Sorry to throw this to you, @droyad, but what do you think about merging this PR?

treesrc:
I think using slash character is better than semicolon. The ability to defining triggers and stored procedure etc. within sql scripts, is a big advantage.

AdrianJSClark:
What will Oracle DB do if it encounters these characters? I know that for SQL Server that the "GO" command isn't recognised by SQL Server itself so we *have* to parse it out. If sending semicolons to Oracle isn't going to change or break the existing behaviour then perhaps we should merge.

Perhaps this PR should also be updated with a note in the breaking changes so end users can be informed?

wazbekker:
I agree with @treesrc. 

This link explains it quite nicely for anyone that is interested: 

[](https://stackoverflow.com/questions/1079949/when-do-i-need-to-use-a-semicolon-vs-a-slash-in-oracle-sql)

droyad:
Right then. Question is, it sounds like a breaking change, with a major version bump (Already!). or is it a bug...

If major, perhaps it would be good to split to a new repo and we start versioning everything independently.

droyad:
Ok, I see one of three solutions:
1) Add an overload that lets you specify the delimiter (non-breaking change and keeping the "incorrect" behaviour
2) Change the default and add an overload that lets you specify the delimiter (breaking change)
3) Forcing a choice of delimiter (breaking change)
4) Mark the old method as deprecated and do 3 (non-breaking change)

treesrc:
@wazbekker I have tried some SQL scripts with /* as a comment inside package creation script. It runs without problem. Could you give sample SQL scripts related to your case? So I can try it.

wazbekker:
@treesrc thanks.  In my case there would be multi line comments above the functions and procedures in the package body.  eg. 

create or replace package body SOME_PACKAGE is
/*---------------------------------------------------
  * Some Description of what the function does
*/---------------------------------------------------
  function some_Function_Call( some_Id in number ) return some_return_type is

treesrc:
@wazbekker the sample script works without any problem: 

```
/* aldkjfaljsdf
*/
create or replace package SOME_PACKAGE1 as
/*---------------------------------------------------
  * Some Description of what the function does
*/---------------------------------------------------
  function some_Function_Call( some_Id in number ) return number;

end;
/
```

But when delimiter and slash star comment is side by side, it does not work (after insert). Is it the expected behavior or not appropriate?

```
/* 
ADFKAL DFAFADF
*/
insert into PRODUCTS (PRODUCT_ID, PRODUCT_NAME, PRODUCT_CATEGORY, PRODUCT_DATE) values (PRODUCT_SEQ.nextval, 'test', 'Q', Current_date)
/  
/* adkfajsdlf
lkjlakj asdf
adsfkljad slfk */
```

wazbekker:
Thanks @treesrc.  

For my use case the delimiter and slash star comment being side by side won't be an issue.  

Any comments from the other guys about this possible issue?

thomasdc:
When will this be merged?

wazbekker:
I found an issue with having the delimiter as a semicolon.

Any use of the divide by arithmetic operator is going to cause a problem.  

A simple example:

`create or replace package some_package is

  procedure divide_two_numbers(p_first_number  in number,
                               p_second_number in number,
                               p_result        out number);

end some_package;
/
create or replace package body some_package is

  procedure divide_two_numbers(p_first_number  in number,
                               p_second_number in number,
                               p_result        out number) is  

begin
  p_result := p_first_number / p_second_number;
end;

end some_package;
`

thomasdc:
[FlyWay](https://flywaydb.org/documentation/database/oracle#example) supports this behavior. They change delimiter based on regex matches. See [here](https://github.com/flyway/flyway/blob/master/flyway-core/src/main/java/org/flywaydb/core/internal/database/oracle/OracleSqlStatementBuilder.java). Maybe we can mimic this behavior?

thomasdc:
> I'm not sure if this is a good idea or not but maybe we could check that the '/' delimiter is followed by "\r\n"?

@wazbekker Line ending behavior is not consistent between Windows (\r\n) and Unix (\n) so I'm not a fan of this approach.

wazbekker:
@thomasdc thanks for confirming its a bad idea.  

Hopefully someone with more DbUp experience can give some input.  @droyad, @AdrianJSClark?

pszittyay:
> @wazbekker Line ending behavior is not consistent between Windows (\r\n) and Unix (\n) so I'm not a fan of this approach.
Why not use Environment.NewLine which in NET CORE is

```
public static String NewLine {
    get {
        Contract.Ensures(Contract.Result() != null);
#if !PLATFORM_UNIX
        return "\r\n";
#else
        return "\n";
#endif // !PLATFORM_UNIX
    }
```
}

https://stackoverflow.com/questions/1015766/difference-between-n-and-environment-newline

dazinator:
So just to confirm, `/` is more like a batch seperator right? In which case the correct analogy to MS SQL would be `GO`.

A semi colon is a line terminator not a batch seperator.

The sql parser allows each provider to specify its own batch seperator to my knowledge.. So for MSSQL provider, this is set to `GO` and there is a lot of parser logic about about ensuring we don't split on GO when its in literals, comments etc etc.

So my thoughts would be, for oracle, you would want to do what the SQL parser for MS SQL does - that is specify a batch seperator of `/` rather than `GO` - I would have thought the existing parser logic which ignores splitting on the batch seperator when found in comments and literals etc would then mostly already take care of what you want?

pszittyay:
@ttufekci could you please change the commit so tha the separator changes to:

`public OracleCommandReader(string sqlText) : base(sqlText, "/"+ Environment.NewLine, delimiterRequiresWhitespace: false)`

nvarscar:
Hi all, any update on this issue? Facing exactly the same problem: DDL statements can't be executed due to semicolon symbol splitting the query.

kjartanp:
I have been using dbUp for oracle now for the last 6 months and it works perfectly, while you are doing basic create table, view , trigger,sequences etc.  But as this issue states it does not work for more complicated things like functions/packages.  To get support for that the separator needs to be "/".  Is there any work being done to get this into next version ?

vethman:
I think I have a better solution. At work we are having a pilot with oracle and DbUp. Too bad that de delimiter is not working for the basic oracle scripts.

As of now I have a working solution with using regex to replace the delimiter. When the pilot is done and it is under test I will make a pull request. Will be my very first one, but I have good developers in my team :)

thomasdcses:
We're using a custom build of dbup-oracle version 4.2.0 with this pull request included for almost a year now and it works great. I wonder why this PR isn't merged yet. 😕

brettveenstra:
That’s good news!  Have you used the transaction options too?  Concerned how that will be supported.



Hacked around this limitation earlier with custom script provider that breaks out “batch1”, “batch2” .. “batchn” but much prefer direct support

sungam3r:
Agree with @thomasdcses. In my opinion, everyone came to the conclusion that it is better (correctly) to use `/` instead of `;`. Is the whole issue reduced to backward compatibility? Who has not used the PL/SQL and continues to not use it, and who used, they still found a workaround. Thus, this is imaginary backward compatibility, the maintenance of which only hurts.

lammaJJ:
Is this pr pending?

p3t3rix:
May i suggest another change to avoid the need to always use / instead of ; ?
What about something like this (without the console outs):

`public IEnumerable<string> SplitScriptIntoCommands(string scriptContents)
{
    using (var reader = new CustomOracleCommandReader(scriptContents))
    {
        var commands = new List<string>();
        reader.ReadAllCommands(c => commands.Add(c));
        var returnCommands = new List<string>();

        foreach (string command in commands)
        {
            if (command.ToLowerInvariant().Contains("begin") || command.ToUpperInvariant().Contains("end"))
            {
                Console.WriteLine($"command with begin/end found, no automatic splitting possible: {command}");
                returnCommands.Add(command);
            }
            else
            {
                returnCommands.AddRange(command.Split(';'));
            }

        }
        return returnCommands;
    }`

This way ; splitting is possible as well as defining inner blocks for trigger etc.

sungam3r:
Well, let's think :)

insert into beginners .... values ..... ;
insert into beginners .... values ..... ;

so "command with begin/end found, no automatic splitting possible" :(

p3t3rix:
yea i realized that afterwards. i guess to really replicate the behavior of sqldeveloper you would have to go full fledged parsing.

I still went for the solution with the begin and end, because the edge cases are easily covered with manual placement of /
So my final adaptation looks like this:

-preprocessor to remove "--"-comments (because somehow they get into the command list)
-preprocessor to add a "/" at the end as a convenience so you don't have to end every file with / if the last command is a trigger etc.
-the adapted command reader as mentioned in this PR
-the splitter as i mentioned above with the additional filter of empty commands (these also throw errors)

Better than the non-working version that is currently checked in and time will tell if we run into more problems 🤞

sungam3r:
So complicated...

Or just change delimiter from ; to /

> Thus, this is imaginary backward compatibility, the maintenance of which only hurts.

rdontireddy:
Hi, is there any plan to approve and merge this PR ?

sungam3r:
@droyad Could you please clarify the situation with this PR? Should there be any other discussions? Do you reject this PR? Are you considering it to merge?

Orkad:
I think the best option for Oracle support could be the possibility to disable the `SqlCommandSplitter` by a `UpgradeEngineBuilder` configuration. The whole file script would be runned by the Oracle engine. What about this ?

Dbup with Oracle is already hard because of unsupported transaction for DDL statement.

hagbourne:
I'm piling on with another +1 for this pull request. After identifying this as the perfect tool for moving to a continuous delivery pipeline but with a legacy Oracle database, my testing halted when I tried to run my first trigger script. 

This pull request resolves the script failures completely. In my case it isn't that I need to split scripts on a forward slash, it is that I don't want it to split on a semicolon. Much as @Orkad suggested. So now I'm running from a locally forked version rather than the published NuGet package until such time that this is merged.

sungam3r:
> So now I'm running from a locally forked version rather than the published NuGet package until such time that this is merged.

Me too 😞

droyad:
Superceeded by #509

dagr8anil:
How does OracleDatabaseWithDefaultDelimiter  handle both a slash to designate a new line and a slash in terms of a math operation?

sungam3r:
`OracleDatabaseWithDefaultDelimiter` work incorrectly with math `/` operations. For those scripts I turned off splitting.

PR #339, opened 2018-05-29
Bug Fix: All non-sql scripts being ignored by .WithScriptsAndCodeEmbeddedInAssembly

joel-cass:
I recently upgraded DbUp from 3.3.5 to 4.1.0. After this, all scripts using IScript were ignored. looking at the code there is a filter applied to WithScriptsAndCodeEmbeddedInAssembly that restricts them to only ending in ".sql", which seems contradictory to the method's intended purpose.

I've removed the default filter with an expression that returns true for all, believing that if a filter is desired it can be entered using the filter parameter.

I think this is somehow related to https://github.com/DbUp/DbUp/issues/213 and its implications didn't quite make it into the documentation.


droyad:
Thanks for the contribution, however this would mean that embedded files are not filtered to just the ones ending in `.sql`. This is fixed in #354

PR #344, opened 2018-07-06
Added SQL Anywhere support.

afurlanet:
Using SQL Anywhere V17 client driver (Sap.Data.SQLAnywhere.v4.5).
Tested against SQLAnywhere 12, 16, 17 servers (all Windows x64).


droyad:
Sorry for the delay. I don't have access to a SQLAnywhere DB, is there a good place to rent one?
I'll merge this since it's a totally new thing and I'll take your word for it that it works :)

PR #348, opened 2018-07-16
Added support for PgSQL client SSL certificates and for DropDatabase

petrce:
Added support for client SSL certificates (used by e.g. Google Cloud SQL) and added support for DropDatabase

### DIFF
--- a/src/dbup-postgresql/PostgresqlExtensions.cs
+++ b/src/dbup-postgresql/PostgresqlExtensions.cs
@@ -6,6 +6,7 @@ using DbUp.Engine.Output;
 using DbUp.Postgresql;
 using DbUp.Engine.Transactions;
 using Npgsql;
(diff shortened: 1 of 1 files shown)


petrce:
hmm, need to find a way how to add certificates without changing the API :/ any ideas?

sabeersulaiman:
@petrce Can we pass this down as a configuration ?

droyad:
You can add overloads instead of adding parameters to existing methods. That'll keep the API compatible.

knunery:
@droyad  I would like to add drop database functionality for postgres. Should I go ahead and make a new PR?

sungam3r:
@petrce Please rebase onto `master` if you are still interested.

droyad:
This was enabled via https://github.com/DbUp/dbup-postgresql/pull/16

PR #350, opened 2018-07-27
Add support for Amazon Redshift

larsbrekken:
Amazon Redshift is based on PostgreSQL and largely supports the same SQL syntax (https://docs.aws.amazon.com/redshift/latest/dg/c_redshift-and-postgres-sql.html).

When trying to use DbUp against Redshift here at [Olo](http://www.olo.com) we have found it to be working well with only a couple of minor adjustments:

* [Redshift does not support `SERIAL`](https://docs.aws.amazon.com/redshift/latest/dg/c_unsupported-postgresql-datatypes.html), so this has been changed to `BIGINT IDENTITY`.
* Redshift does not have a built-in database named `postgres`, so the check for whether the database already existed was failing. Both Redshift and PostgreSQL have default databases called [`template1`](https://www.postgresql.org/docs/current/static/manage-ag-templatedbs.html), so by using this database instead of `postgres` we get support for both.

We considered adding a separate library for Redshift instead of modifying PostgreSQL, however with them being 99% identical this seemed like overkill.

The downside is the number of times the default parameter `isAmazonRedshift` had to be added to `PostgresqlExtensions`, however since it's by default set to `false` the change should be backwards compatible and opt-in for consumers using Redshift instead of regular PostgreSQL.


larsbrekken:
I also added a commit for running Assent as non-interactive so that it shows the specific error instead of failing when trying to launch a diff tool (see https://github.com/droyad/Assent/blob/master/README.md#automated-builds).

image

droyad:
Is it possible the user has deleted `template1`?

larsbrekken:
@droyad Based on https://docs.aws.amazon.com/redshift/latest/dg/r_DROP_DATABASE.html it does not seem possible to delete `template1` in Redshift. 

Based on https://www.postgresql.org/docs/current/manage-ag-templatedbs.html it seems theoretically possible to delete it from Postgres after setting `datistemplate = false`, but from what I gather this seems rather unlikely in practice.

droyad:
Sorry about the delay in getting back to this. I'd prefer to keep backwards compatibility outside a `major` bump, so if it's possible someone has deleted the template this would be a breaking change. Perhaps we need to just change the from `master` to `template1` if it's redshift.

Also the tests are failing because the API has changed. Adding an parameter to a method is a breaking change (optional parameters are compiler sugar). Please add an overload instead.

larsbrekken:
Thanks Robert, that makes sense. Let me work on addressing your feedback there.

larsbrekken:
@droyad Adding overloads to the Postgres project was getting pretty messy, so I instead opted to create a separate Redshift project. This should make it easier to handle any additional changes between the two databases in the future, with the cost of some code duplication.

If this seems like an appropriate solution to you, I will do some testing against a Redshift database.

larsbrekken:
I've done some testing against a Redshift database now, verifying that a script can be applied successfully, plus some error cases (insufficient permissions to `schemaversions`, invalid syntax in SQL script).

droyad:
Fixes #352

Issue #159, opened 2016-01-26
Firebird upgrade script with multiple commands 


petteri-laine:
Firebird sql upgrade script is failing if it has multiple statements.

```
/* header comment above the statments */
ALTER TABLE "MyTable" ADD "foo" int default null;
ALTER TABLE "MyTable" ADD "bar" int default null;
```

Exception points to the second statement of the script.

Script block number: 0; Error code 335544569; Message: Dynamic SQL Error
SQL error code = -104
Token unknown - line 9, column 1
ALTER
2016-01-25 16:19:02.5423 ERROR FirebirdSql.Data.FirebirdClient.FbException (0x80004005): Dynamic SQL Error
SQL error code = -104
Token unknown - line 9, column 1
ALTER ---> Dynamic SQL Error
SQL error code = -104
Token unknown - line 9, column 1
ALTER
   at FirebirdSql.Data.FirebirdClient.FbCommand.ExecuteNonQuery() in c:\path\NETProvider\working\NETProvider\src\FirebirdSql.Data.FirebirdClient\FirebirdClient\FbCommand.cs:line 504
   at DbUp.Support.SqlServer.SqlScriptExecutor.<>c__DisplayClass15_0.<Execute>b__1(Func`1 dbCommandFactory) in C:\projects\dbup\src\DbUp\Support\SqlServer\SqlScriptExecutor.cs:line 125
   at DbUp.Engine.Transactions.TransactionPerScriptStrategy.Execute(Action`1 action) in C:\path\dbup\src\DbUp\Engine\Transactions\TransactionPerScriptStrategy.cs:line 16
   at DbUp.Engine.Transactions.DatabaseConnectionManager.ExecuteCommandsWithManagedConnection(Action`1 action) in C:\path\dbup\src\DbUp\Engine\Transactions\DatabaseConnectionManager.cs:line 100
   at DbUp.Support.SqlServer.SqlScriptExecutor.Execute(SqlScript script, IDictionary`2 variables) in C:\path\dbup\src\DbUp\Support\SqlServer\SqlScriptExecutor.cs:line 106

dazinator:
Guessing Firebird ado.net provider (or fb itself) has an issue with doing multiple `Alter`'s within the same batch / `DbCommand`. To test that theory, you can try seperating those 2 alter statements into 2 seperate scripts. 

Also if fb permits adding multiple columns in a single alter statement then you might try that..

dazinator:
Looked into this a bit further.. 

[...]

PR #365, opened 2018-09-21
Create/Drop Firebird databases (issue #362)

hhindriks:
Implements DropDatabase and EnsureDatabase in the dpup-firebird project #362.
`DropDatabase.For.FirebirdDatabase(connectionString);
EnsureDatabase.For.FirebirdDatabase(connectionString);`

Replaced the code copied from Postgres to split a script with a Firebird specific one as mentioned in #159.

Use unquoted table name for schemaversions table, in the generator and the trigger, they were inconsistent before.

`DoesTableExistSql()` fixed (see #248 and #250). Firebird uses all uppercase for database objects unless they are quoted. Per above changed to unquoted names and using an UPPER(tablename) in the DoesTableExistSQL().

### DIFF
--- a/src/dbup-firebird/FirebirdConnectionManager.cs
+++ b/src/dbup-firebird/FirebirdConnectionManager.cs
@@ -1,8 +1,7 @@
 ﻿using DbUp.Engine.Transactions;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text.RegularExpressions;
 using FirebirdSql.Data.FirebirdClient;
+using FirebirdSql.Data.Isql;
+using System.Collections.Generic;
 
 namespace DbUp.Firebird
 {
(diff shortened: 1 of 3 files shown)


hhindriks:
Failed to mention: Tested against Firebird 3.0.3 both in embedded mode and remotly.

Had to increase the FirebirdClient to v.5.0.0.

droyad:
The `FirebirdClient` bump means it's technically a major version bump.. but it's unlikely to be an issue for the majority of users. @jburger thoughts?

This might be a good case for splitting the providers into separate repos, then they can be versioned at different paces.

sungam3r:
@hhindriks Please rebase onto `master` if you are still interested.

src/dbup-firebird/FirebirdConnectionManager.cs
Jeern, line 1: This is actually a good idea to use the FbScript in the firebird provider. However it is not backwards compatible. E.g.. this does not work which is bad: 

CREATE PROCEDURE FOO
AS
BEGIN
  EXIT;
END

On the other hand this works:

SET TERM ^ ;

CREATE PROCEDURE FOO
AS
BEGIN
  EXIT;
END^

SET TERM ; ^

Which is great because it does not work with the existing implementation.

Still I cannot approve that it is not backward compatible, We could consider making it optional (i.e. a classic vs. a FbScript mode)

mjauernig: How should the medium-term implementation look like? Drop _classic_ or co-exist both versions? 

If _drop-classic_ then we can mark it in version **x** as deprecated and remove it in version **x+1**.

mjauernig: Ok, both versions are fine.

droyad:
See https://github.com/DbUp/DbUp/issues/773 on how to move this to the `dbup-firebird` repo

Issue #368, opened 2018-09-28
Add Extensions to FileSystemScriptOptions


lvermeulen:
Currently the only extension that is considered by DbUp's FileSystemScriptProvider is *.sql. 

In FileSystemScriptProvider.GetScripts:
            var files = Directory.GetFiles(directoryPath, "*.sql", ShouldSearchSubDirectories()).AsEnumerable();

I request the addition of an Extensions property to FileSystemScriptOptions, as we have a need to consider a couple of other extensions as well. A PR will be sent.

PR #369, opened 2018-09-28
Add Extensions to FileSystemScriptOptions

lvermeulen:
Fixes https://github.com/DbUp/DbUp/issues/368

### DIFF
--- a/src/dbup-core/ScriptProviders/FileSystemScriptOptions.cs
+++ b/src/dbup-core/ScriptProviders/FileSystemScriptOptions.cs
@@ -14,6 +14,7 @@ namespace DbUp.ScriptProviders
         public FileSystemScriptOptions()
         {
             Encoding = DbUpDefaults.DefaultEncoding;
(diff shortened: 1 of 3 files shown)


src/dbup-core/ScriptProviders/FileSystemScriptOptions.cs
```diff
@@ -30,5 +31,10 @@ public FileSystemScriptOptions()
         /// The encoding to be used for reading files 
         /// </summary>
         public Encoding Encoding { get; set; }
+
+        /// <summary>
+        /// The SQL upgrade script extensions
+        /// </summary>
+        public string[] Extensions { get; set; }
```
sungam3r: NoPublicApiChanges test failed

droyad: Ah thanks, we were zooming through the Issues/PRs today

PR #374, opened 2018-10-17
Better parsing of double quoted texts.

### LINKED ISSUE #351: Combination of double quotes, asterisks, and GO command breaks deployment
If a script both selects a column as a name using double quotes (i.e. SELECT 1 AS "Test/*") and uses a GO command, the upgrade process will fail with message 'Incorrect syntax near GO'. Removing the GO command, or converting the double quotes to either single quotes or square brackets fixes the issue.

Tested on SqlServer 2016 and 2017.

Code to reproduce:

static void Main(string[] args)
        {
            var connectionString = ConfigurationManager.ConnectionStrings["ConnectionString"].ConnectionString;

            /* 
             * SELECT 'Test' AS "Test/*"
             * GO
             */
            var sql = "SELECT 'Test' AS \"Test/*\"\nGO";
            var script = new DbUp.Engine.SqlScript("Tester", sql);

            var builder = DbUp.DeployChanges.To
                .SqlDatabase(connectionString)
                .JournalTo(new NullJournal())
                .WithScripts(script)
                .WithTransaction()
                .LogToConsole()
                .Build();

            var result = builder.PerformUpgrade();
            Debug.Assert(!result.Successful); // Fails

            /*
             * SELECT 'Test' AS "Test/*"
             */
            sql = "SELECT 'Test' AS \"Test/*\"";
            script = new DbUp.Engine.SqlScript("Tester", sql);

            builder = DbUp.DeployChanges.To
            .SqlDatabase(connectionString)
            .JournalTo(new NullJournal())
            .WithScripts(script)
            .WithTransaction()
            .LogToConsole()
            .Build();

            result = builder.PerformUpgrade();
            Debug.Assert(result.Successful); // Succeeds

            /* 
             * SELECT 'Test' AS 'Test/*'
             * GO
             */
            sql = "SELECT 'Test' AS 'Test/*'\nGO";
            script = new DbUp.Engine.SqlScript("Tester", sql);

            builder = DbUp.DeployChanges.To
                .SqlDatabase(connectionString)
                .WithScripts(script)
                .JournalTo(new NullJournal())
                .WithTransaction()
                .LogToConsole()
                .Build();

            result = builder.PerformUpgrade();
            Debug.Assert(result.Successful); // Succeeds
        }
    }

sgoodgrove:
This should resolve #351 by supporting double quoted strings in the same manner as single quoted strings. It may be that IsDoubleQuote should be a protected member, but that throws out the API tests for changes to the public signature.

### DIFF
--- a/src/dbup-core/Support/SqlParser.cs
+++ b/src/dbup-core/Support/SqlParser.cs
@@ -12,6 +12,7 @@ namespace DbUp.Support
         private const char EndOfLineChar = '\n';
         private const char CarriageReturn = '\r';
         private const char SingleQuoteChar = (char)39;
+        private const char DoubleQuoteChar = '"';
         private const char DashChar = '-';
         private const char SlashChar = '/';
         private const char StarChar = '*';
@@ -66,7 +67,7 @@ namespace DbUp.Support
                 {
                     ReadCustomStatement();
                 }
-                else if (IsQuote)
+                else if (IsQuote || IsDoubleQuote)
                 {
                     ReadQuotedString();
                 }
@@ -290,6 +291,14 @@ namespace DbUp.Support
             return true;
         }
 
+        private bool IsDoubleQuote
+        {
+            get
+            {
+                return CurrentChar == DoubleQuoteChar;
+            }
+        }
+
         private bool IsBeginningOfBracketedText
         {
             get
@@ -355,6 +364,7 @@ namespace DbUp.Support
         /// </summary>
         private void ReadQuotedString()
         {
+            var quoteChar = CurrentChar;
             ReadCharacter(CharacterType.QuotedString, CurrentChar);
             while (Read() != FailedRead)
             {
@@ -364,7 +374,7 @@ namespace DbUp.Support
                     Read();
                 }
                 ReadCharacter(CharacterType.QuotedString, CurrentChar);
-                if (IsQuote)
+                if (CurrentChar == quoteChar)
                 {
                     return;
                 }
@@ -514,7 +524,7 @@ namespace DbUp.Support
             Delimiter,
             /// <summary>Character is a custom statement (open for new implementation)</summary>
             CustomStatement,
-            
+
         }
     }
 }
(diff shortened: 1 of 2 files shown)


droyad:
@jburger or @AdrianJSClark Could you have a quick look at this? From what I can see it shouldn't break anything, but this parser always make me nervous.

src/dbup-core/Support/SqlParser.cs
```diff
@@ -355,6 +364,7 @@ private bool IsEndOfSlashStarComment
         /// </summary>
         private void ReadQuotedString()
         {
+            var quoteChar = CurrentChar;
```
AdrianJSClark:
```suggestion
            var currentQuoteChar = CurrentChar;
```

src/dbup-core/Support/SqlParser.cs
```diff
@@ -364,7 +374,7 @@ private void ReadQuotedString()
                     Read();
                 }
                 ReadCharacter(CharacterType.QuotedString, CurrentChar);
-                if (IsQuote)
+                if (CurrentChar == quoteChar)
```
AdrianJSClark:
```suggestion
                if (CurrentChar == currentQuoteChar)
```

src/dbup-core/Support/SqlParser.cs
AdrianJSClark: This variable naming makes more sense to me, especially further down where it is used. YMMV - happy to leave it as-is.

src/dbup-tests/Engine/VariableSubstitutionPreprocessorTests.cs
```diff
@@ -35,6 +35,27 @@ public void substitutes_variables_in_body()
             command.CommandText.ShouldBe("something coriander something");
         }
 
+        [Fact]
+        public void substitutes_variables_when_quoted_text_contains_comment_sequence()
```
AdrianJSClark: I don't have time to look right this second, but are there other tests for the single-quote parsing that should be duplicated now we are parsing double-quotes as well?

I'm thinking that I'm sure there are checks to ensure the parser correctly quote-matches, but I don't see them being updated in this PR.

sungam3r:
@sgoodgrove Please rebase onto `master` if you are still interested.

sgoodgrove:
@sungam3r this is not a huge issue for me, but happy to update.

PR #380, opened 2018-10-26
Prevents interactive reporting for assent

jburger:
Currently our builds are failing due to Assent attempting to spin up a reporting tool like beyond compare or KDiff. 

Looking at this: https://github.com/droyad/Assent#automated-builds looks like setting an environment variable should allow builds to proceed.


jburger:
Tested build here - seems to work OK https://ci.appveyor.com/project/getpack/dbup/build/tests

PR #382, opened 2018-10-30
Allow specific collation on db creation (sql server)

tiptobi:


src/dbup-sqlserver/SqlServerExtensions.cs
```diff
@@ -191,10 +250,12 @@ public static void SqlDatabase(this SupportedDatabasesForEnsureDatabase supporte
                 }
             }
 
+            string collationString = String.IsNullOrEmpty(collation) ? "" : string.Format(@" COLLATE {0}", collation);
```
AdrianJSClark: Minor nitpick to be consistent with referencing `string`:

```suggestion
            string collationString = string.IsNullOrEmpty(collation) ? "" : string.Format(@" COLLATE {0}", collation);
```

src/dbup-sqlserver/SqlServerExtensions.cs
```diff
@@ -135,6 +135,18 @@ public static void SqlDatabase(this SupportedDatabasesForEnsureDatabase supporte
     {
         SqlDatabase(supported, connectionString, new ConsoleUpgradeLog(), commandTimeout);
     }
+    
+    /// <summary>
+    /// Ensures that the database specified in the connection string exists.
+    /// </summary>
+    /// <param name="supported">Fluent helper type.</param>
+    /// <param name="connectionString">The connection string.</param>
+    /// <param name="collation">The collation of the database to create</param>
```
AdrianJSClark:
```suggestion
    /// <param name="collation">The collation name to set during database creation</param>
```

PR #400, opened 2019-03-25
Semi-colon for Azure SqlDatabase

### LINKED ISSUE #393: EnsureDatabase for Azure SQL fails
Running `EnsureDatabase.For.SqlDatabase()` against Azure SQL database fails because of invalid command constructed by `SqlServerExtensions.SqlDatabase(this SupportedDatabasesForEnsureDatabase supported, string connectionString, IUpgradeLog logger, int timeout = -1, AzureDatabaseEdition azureDatabaseEdition = AzureDatabaseEdition.None, string collation = null)`. The problem seems to be:

`sqlCommandText = string.Format
(
  @"create database [{0}]{1};",
  databaseName,
  collationString
);`

where I can see semicolon being added in command text. However, it is followed by switch for Azure DB edition. So, the result command text I can see is:

`create database [DATABASE_NAME]; ( EDITION = ''basic'' );`

And it fails with:

`Incorrect syntax near 'EDITION'.'`

Thanks.

sungam3r:
fixes #393

### DIFF
--- a/src/dbup-sqlserver/SqlServerExtensions.cs
+++ b/src/dbup-sqlserver/SqlServerExtensions.cs
@@ -253,13 +253,16 @@ public static class SqlServerExtensions
             string collationString = string.IsNullOrEmpty(collation) ? "" : string.Format(@" COLLATE {0}", collation);
             sqlCommandText = string.Format
                     (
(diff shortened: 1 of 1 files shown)


sungam3r:
Any progress on this?

PR #403, opened 2019-04-02
Update variable-substitution.md

istvan2015:
Error in documentation - variables must be enclosed in $ signs (2 of them)

### DIFF
--- a/docs/more-info/variable-substitution.md
+++ b/docs/more-info/variable-substitution.md
@@ -6,17 +6,31 @@ DeployChanges.To
   .WithVariable("TestVariable", "Value")
 ```
 
-Then in your database script:
+For more than one variable, use a Dictionary:
+``` csharp
+var myVariables = new Dictionary<string, string>
+{
+  {"Variable1", "Value1"},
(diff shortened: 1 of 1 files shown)


docs/more-info/variable-substitution.md
```diff
@@ -6,17 +6,31 @@ DeployChanges.To
...
 -- $TestVariable$ $AnotherVariable$
 print '$TestVariable$'
-SELECT * FROM dbo.$TestVariable
+SELECT * FROM dbo.$TestVariable$
 ```
 
-Will execute:
+Will result in:
 ```
--- Value $AnotherVariable$
```
droyad: I don't think this should have been removed

istvan2015: I removed it because above that, it was already made clear that the variable must be enclosed in two $ signs. In there, where I removed it, I showed the example of what the above may translate into - and it won't translate keeping the enclosing $...$ pair, if that makes sense. Hence it was removed, and it makes perfect sense to be removed, I believe. What do you think, @droyad ?

PR #415, opened 2019-05-30
Add event for script execution

### LINKED ISSUE #390: AdHocSqlRunner.ExecuteReader fails when output columns have no names
Any attempt to run an adhoc query with an output that has 2 or more columns without defined names fails with the following exception:
```
Exception calling "ExecuteReader" with "1" argument(s): "An item with the same key has already been added. Key: "
```
Since it returns dictionaries, it kind of makes sense, yet I cannot workaround that by assigning column names manually since they are defined in the DbUp code. Any chance those names could be generated automatically if missing? Like, Column1, Column2, etc. Other option would probably be to build `[System.Data.DataTable]` object instead of a set of dictionaries.

Test code (I totally realize this might not be an intended way to run AdHoc code, but that's all I have right now):
```powershell
$connectionString = 'Data Source=SQL1;User=sqldeploy;Password=sqldeploy;'
$dbUpConnection = [DbUp.SqlServer.SqlConnectionManager]::new($connectionString)
$dbUpSqlParser = [DbUp.SqlServer.SqlServerObjectParser]::new()
$managedConnection = $dbUpConnection.OperationStarting($null, $null)
$dbUpConnection.ExecuteCommandsWithManagedConnection( [Func[Func[Data.IDbCommand], System.Object]] {
    Param (
        $dbCommandFactory
    )
    $sqlRunner = [DbUp.Helpers.AdHocSqlRunner]::new($dbCommandFactory, $dbUpSqlParser, $null)
    $sqlRunner.ExecuteReader('SELECT 1, 2')
})
$managedConnection.Dispose()
```

Tested on v4.2.0.

sungam3r:
resurrected #241
also fixes #390

### DIFF
--- a/src/dbup-core/Engine/ScriptExecutedEventArgs.cs
+++ b/src/dbup-core/Engine/ScriptExecutedEventArgs.cs
@@ -1,0 +1,27 @@
+﻿using DbUp.Engine.Transactions;
+using System;
+
+namespace DbUp.Engine
+{
+    /// <summary>
+    /// Provides the information for the most recently executed script.
+    /// </summary>
+    public class ScriptExecutedEventArgs : EventArgs
+    {
+        public ScriptExecutedEventArgs(SqlScript script, IConnectionManager connectionManager)
+        {
+            Script = script;
+            ConnectionManager = connectionManager;
+        }
+
+        /// <summary>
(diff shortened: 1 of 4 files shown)


sungam3r:
@droyad Any objections?

PR #435, opened 2019-09-13
Added JournalToMySqlTable for journaling to MySql

slamballx:


sungam3r:
@slamballx Please rebase onto `master` if you are still interested.

slamballx:
https://github.com/sungam3r I think this commit is no more useful. I can see that now esists a specific class for journaling on MySQL https://github.com/DbUp/DbUp/blob/master/src/dbup-mysql/MySqlTableJournal.cs.

I close this. Thanks

sungam3r:
ok

PR #440, opened 2019-10-05
DbCommandFactory: create DbCommand with committed transaction

### LINKED ISSUE #437: TransactionPerScriptStrategy: Create DbCommand with committed transaction
For reproduce this issue we can use script from samples

https://github.com/DbUp/DbUp/blob/master/src/Samples/SampleApplication/Scripts/Script0005%20-%20ComplexUpdate.cs#L17

if we use `commandFactory` for create dbCommand  :

```
public string ProvideScript(Func<IDbCommand> commandFactory)
{
    var dbcommand = commandFactory();  // DbCommand is creat with committed transaction
         dbcommand .CommandText = "select 1"; 
         dbcommand .ExecuteScalar();
}
```

It creates `DbCommand `  with committed transaction because it does call back to result after  `transaction.Commit();`  has been already invoked :

https://github.com/DbUp/DbUp/blob/e874234e8d320dc37c272e5137092da0b89d72f3/src/dbup-core/Engine/Transactions/TransactionPerScriptStrategy.cs#L30-L37

I couldn't find in documentation if we should use `commandFactory` to created `dbcommand` inside `ProvideScript` method from `IScript`.

Created `DbCommand ` this way in `SQLite` flavour cause exception during DbCommand execution because assigned transaction is committed.

liborweigl:
This should fix  DbUp/DbUp#801 bug with committed transaction for LazySqlScript when `Dbcommand` is created through ` DbCommandFactory` inside code scripts.

### DIFF
--- a/src/dbup-core/ScriptProviders/EmbeddedScriptAndCodeProvider.cs
+++ b/src/dbup-core/ScriptProviders/EmbeddedScriptAndCodeProvider.cs
@@ -66,18 +66,18 @@ public class EmbeddedScriptAndCodeProvider : IScriptProvider
     IEnumerable<SqlScript> ScriptsFromScriptClasses(IConnectionManager connectionManager)
     {
         var script = typeof(IScript);
-            return assembly
+        return  assembly
             .GetTypes()
             .Where(type => script.IsAssignableFrom(type) &&
                            type.IsClass &&
                            !type.IsAbstract
             )
-                .Select(s => (SqlScript)new LazySqlScript(s.FullName + ".cs", sqlScriptOptions, () =>
+            .Select(s => (SqlScript)new LazySqlScript(s.FullName + ".cs", sqlScriptOptions, () =>
                 connectionManager.ExecuteCommandsWithManagedConnection(
                     dbCommandFactory =>
-                                        ((IScript)Activator.CreateInstance(s))
-                                        .ProvideScript(dbCommandFactory)))
-                ).ToList();
+                        ((IScript)Activator.CreateInstance(s))
+                        .ProvideScript(dbCommandFactory)))
+            ).ToList();
     }
 
     /// <summary>
--- a/src/dbup-core/ScriptProviders/ScriptInstanceProvider.cs
+++ b/src/dbup-core/ScriptProviders/ScriptInstanceProvider.cs
@@ -26,7 +26,8 @@ class ScriptInstanceProvider : IScriptProvider
     /// </summary>
     /// <param name="scripts">The IScript instances to include</param>
     /// <param name="namer">A function that returns the name of the script</param>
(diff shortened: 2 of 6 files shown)


src/dbup-core/ScriptProviders/EmbeddedScriptAndCodeProvider.cs
```diff
@@ -66,7 +66,7 @@ public EmbeddedScriptAndCodeProvider(Assembly assembly, Func<string, bool> filte
         IEnumerable<SqlScript> ScriptsFromScriptClasses(IConnectionManager connectionManager)
         {
             var script = typeof(IScript);
-            return connectionManager.ExecuteCommandsWithManagedConnection(dbCommandFactory => assembly
```
liborweigl: This cause that transaction has been committed before create Dbcommands for generating content in `LazySqlScript`

src/dbup-core/ScriptProviders/EmbeddedScriptAndCodeProvider.cs
```diff
@@ -79,8 +79,12 @@ IEnumerable<SqlScript> ScriptsFromScriptClasses(IConnectionManager connectionMan
...
 #endif
                 })
-                .Select(s => (SqlScript)new LazySqlScript(s.FullName + ".cs", sqlScriptOptions, () => ((IScript)Activator.CreateInstance(s)).ProvideScript(dbCommandFactory)))
-                .ToList());
+                .Select(s => (SqlScript)new LazySqlScript(s.FullName + ".cs", sqlScriptOptions, () =>
+                connectionManager.ExecuteCommandsWithManagedConnection(
+                    dbCommandFactory =>
+                                        ((IScript)Activator.CreateInstance(s))
+                                        .ProvideScript(dbCommandFactory)))
+                ).ToList();
```
liborweigl: It will create `Transaction` per `LazySqlScript`

droyad:
I'm pondering whether to put this into the 6.0 release.

I haven't had a chance to look deeply into this problem yet. On first brush the thing giving me pause is that the with this fix the `ProvideScript` method runs in a different transaction to the script that is returned. (I think)

I'm not sure if that's a problem or not, need to mull on it.

Also in the wild I've seen a common pattern where the migration is done in the body of the `ProvideScript` method and then returns an empty string.

mikkelbu:
> Also in the wild I've seen a common pattern where the migration is done in the body of the `ProvideScript` method and then returns an empty string.

Isn't this the pattern shown in the end of https://dbup.readthedocs.io/en/latest/usage/ - in `Script0006UpdateInCode` ?

droyad:
As I suspected this change would cause the ProvideScript and the script returned from the method to run under different transactions. 

However considering that the app should be offline when migration runs, I think that's ok. It's unlikely that both sets of scripts modify the data.

In any case, this makes it better. The ProvideScript method not runs under a transaction when using TransactionPerScript when it didn't previously.

PR #443, opened 2019-10-08
Fix Nested Multiline Comment Parsing by SqlParser

### LINKED ISSUE #439: [DbUp-Core] SQLParser fails to parse some nested block comments
this fails (no space between end block comment chars):
`/* comment1 /* comment2 */*/`

this don't:
`/* comment1 /* comment2 */ */`

AdrianJSClark:
When the SqlParser encountered a nested multiline comment that had the
closing comment delimeters together in sequence it would fail to
properly parse.

Given the example:
```sql
    /*ab/*cd*/*/
```
The SqlParser correctly identifies the second multiline comment
(i.e. `cd`) it then advances until the `/` character which closes the
inner comment (10th character in the string). At this character, the
parser properties were looking forward and seeing `/*` but also looking
backwards and seeing `*/`. Because the check for the start of another
nested comment was checked first it would be read as the start of
another multiline comment.

Similar to the nested multiline comments where closing tokens came
one-after-the-other, the opening tokens could also be confused.

This PR resolves the issue by changing the end multiline comment to
be detected by looking forward. Then, when either the start or end
of the multiline comment is detected both characters of the token are
immediately processed.

Fixes #439

### DIFF
--- a/src/dbup-core/Support/SqlParser.cs
+++ b/src/dbup-core/Support/SqlParser.cs
@@ -363,8 +363,12 @@ namespace DbUp.Support
 
         void ReadSlashStarComment()
         {
-            // Write the current slash.
+            // We have entered this method because we've identified the "/*" pattern.
+            // Write both characters here as technically they go together as a token.
             ReadCharacter(CharacterType.SlashStarComment, CurrentChar);
+            Read();
+            ReadCharacter(CharacterType.SlashStarComment, CurrentChar);
+
             // Read until we find a the ending of the slash star comment,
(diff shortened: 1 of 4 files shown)


jarnoj:
@AdrianJSClark Perhaps someone else could review this commit as @droyad might be less active around here..?

AdrianJSClark:
Hi @jarnoj, sorry about the delay here. I'll merge this since I'm pretty sure of the fix and the fact that I've included tests means there's unlikely to be any huge issues.

Issue #426, opened 2019-08-16
Add a way to get UpgradeConfiguration from UpgradeEngineBuilder


joaoportela:
In more complex usage scenarios (that aren't permitted by UpgradeEngine directly) I need to create an UpgradeConfiguration (or some of its parts) manually.

If I could get the UpgradeConfiguration directly from UpgradeEngineBuilder I could still configure everything using the handy fluent interface.

droyad:
Apologies for the delay @joaoportela. The `UpgradeEngineBuilder.Configure` method is what you are looking for. It's what all the other configuration methods use in the end.

joaoportela:
That would be great! 😃

joaoportela:
I'm going to try and do this real quick. PR comming soon. :)

PR #459, opened 2019-10-26
Allow getting UpgradeConfiguration from UpgradeEngineBuilder

joaoportela:
Add a new method called BuildConfiguration that gets
theUpgradeConfiguration that would have been used to create
theUpgradeEngine.

Closes DbUp/DbUp#426


joaoportela:
I didn't write any tests for this because I wasn't sure how to do it.

Any suggestions?

droyad:
Looks good to me, not worth adding extra tests.

PR #465, opened 2019-11-21
Allow use of Azure AppAuthentication integrated security

justinjstark:
Adds an optional flag to SqlServer to UseAzureIntegratedSecurity. This allows us to use the new AppAuthentication package which connects to Azure resources using integrated security from within a Visual Studio instance or from a Managed Identity for a service on Azure.

### DIFF
--- a/src/dbup-sqlserver/SqlConnectionManager.cs
+++ b/src/dbup-sqlserver/SqlConnectionManager.cs
@@ -3,6 +3,9 @@ using System.Collections.Generic;
 using System.Data.SqlClient;
 using DbUp.Engine.Transactions;
 using DbUp.Support;
+#if SUPPORTS_AZURE_AD
+using Microsoft.Azure.Services.AppAuthentication;
+#endif
 
 namespace DbUp.SqlServer
 {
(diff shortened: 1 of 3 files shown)


unicrus:
Why not just use Microsoft.Data.SqlClient? https://github.com/dotnet/SqlClient
It supports all kinds of Azure authentication.

AdrianJSClark:
@unicrus That's a large change which would require a major version bump. It is something that is under consideration for inclusion before the next time we do a major release like that.

unicrus:
@AdrianJSClark why is it major?
The only change is a replacement of System.Data.SqlClient.SqlConnection with Microsoft.Data.SqlClient.

Currently, I've had to do my own SqlConnectionManager to have Azure AD password authentication. And it has just one line different from the original SqlConnectionManager.

AdrianJSClark:
@unicrus We can't change out the library which supports the major functionality of the DbUp SqlServer package without triggering a major version change. The work to integrate the new library may be small, but what is being swapped out is fundamental. End users of the library need that change communicated.

We don't know the interesting ways downstream consumers of the package might have woven dependencies on both System.Data.SqlClient and DbUp into their own code. Changing to use Microsoft.Data.SqlClient could possibly cause code like that to break. That's why the major version number change is the right thing to do.

PR #466, opened 2019-11-22
Make useAzureSqlIntegratedSecurity optional & fix target framework msg

justinjstark:
A couple fixes to yesterday's merged PR #465.

1) Make SqlConnectionManager:useAzureIntegratedSecurity parameter optional. Before this fix, it was a breaking change to anyone using SqlConnectionManager directly.
2) Correct the minimum supported frameworks in the unsupported-framework exception message.

### DIFF
--- a/src/dbup-sqlserver/SqlConnectionManager.cs
+++ b/src/dbup-sqlserver/SqlConnectionManager.cs
@@ -18,7 +18,7 @@ namespace DbUp.SqlServer
         /// Manages Sql Database Connections
         /// </summary>
         /// <param name="connectionString"></param>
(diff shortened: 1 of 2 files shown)


sungam3r:
> it was a breaking change to anyone using SqlConnectionManager directly

Any new parameter (even ones with default values) in public api is a breaking change - client code must be recompiled in order to use new method signature.

sungam3r:
Keeping this in mind, it is more profitable for libraries to provide a second method with an extended signature and simply reuse the implementation.

justinjstark:
@sungam3r Those are good points. With that in mind, I've updated the code to use an overload and keep the original ctor signature the same. I'll let the DbUp team choose to include this or scrap it.

PR #470, opened 2019-11-29
Update to Microsoft.Data.SqlClient

johnsimons:
This would be a breaking change.


sungam3r:
You can use the new package without breaking backward compatibility - it is enough to use preprocessor directives for usings. No need to change the list of target platforms.

johnsimons:
No longer needed, found a different way

michaeltnguyen:
@johnsimons how did you end up doing this?

johnsimons:
Ended up creating https://github.com/OctopusDeploy/Octopus.dbup.SqlServer

PR #474, opened 2019-12-09
Enable using SSL with Postgres

ken-wolfe:
New PostgresqlConnectionManager that takes a byte array
of the .pfx file to be used as the certificate for connecting
to Postgres using SSL.

New PostgresqlExtension method for creating an
UpgradeEngineBuilder with the certificate.

Minor refactor of EnsureDatabase function so that it
can optionally take the certificate.

### DIFF
--- a/src/dbup-postgresql/PostgresqlConnectionManager.cs
+++ b/src/dbup-postgresql/PostgresqlConnectionManager.cs
@@ -25,13 +25,13 @@ namespace DbUp.Postgresql
         /// Creates a new PostgreSQL database connection with a certificate.
         /// </summary>
         /// <param name="connectionString">The PostgreSQL connection string.</param>
-        /// <param name="certificateFile">A byte array representation of an X509 certificate.</param>
-        public PostgresqlConnectionManager(string connectionString, byte[] certificateFile)
+        /// <param name="certificate">Certificate for securing connection.</param>
+        public PostgresqlConnectionManager(string connectionString, X509Certificate2 certificate)
             : base(new DelegateConnectionFactory(l =>
                 {
                     NpgsqlConnection databaseConnection = new NpgsqlConnection(connectionString);
                     databaseConnection.ProvideClientCertificatesCallback +=
-                        certs => certs.Add(new X509Certificate2(certificateFile));
+                        certs => certs.Add(certificate);
 
                     return databaseConnection;
                 }
--- a/src/dbup-postgresql/PostgresqlConnectionManager.cs
+++ b/src/dbup-postgresql/PostgresqlConnectionManager.cs
@@ -1,7 +1,7 @@
 ﻿using System.Collections.Generic;
 using System.Linq;
-using System.Text.RegularExpressions; 
 using System.Security.Cryptography.X509Certificates;
+using System.Text.RegularExpressions; 
 using DbUp.Engine.Transactions;
 using Npgsql;
 
--- a/src/dbup-postgresql/PostgresqlConnectionManager.cs
+++ b/src/dbup-postgresql/PostgresqlConnectionManager.cs
@@ -1,6 +1,7 @@
 ﻿using System.Collections.Generic;
 using System.Linq;
-using System.Text.RegularExpressions;
+using System.Text.RegularExpressions; 
+using System.Security.Cryptography.X509Certificates;
 using DbUp.Engine.Transactions;
 using Npgsql;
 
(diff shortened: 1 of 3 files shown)


ken-wolfe:
This is my first contribution, so hopefully I did the right things. Please let me know what you'd like fixed or any comments that you might have. Happy to have the feedback.

ken-wolfe:
I tested this on a database that we have in Google Cloud with the certificates generated in the cloud console. Then I used openssl to generate the .pfx file that I read in and passed to the new functions that I created.

src/dbup-postgresql/PostgresqlConnectionManager.cs
```diff
@@ -20,6 +21,24 @@ public PostgresqlConnectionManager(string connectionString)
         {
         }
 
+        /// <summary>
+        /// Creates a new PostgreSQL database connection with a certificate.
+        /// </summary>
+        /// <param name="connectionString">The PostgreSQL connection string.</param>
+        /// <param name="certificateFile">A byte array representation of an X509 certificate.</param>
+        public PostgresqlConnectionManager(string connectionString, byte[] certificateFile)
```
AdrianJSClark: Is there a particular reason the external interfaces use `byte[]` rather than the type that the `Add(...)` method of the `X509CertificateCollection` class expects, which is something that derives from `X509Certificate`? (See [X509CertificateCollection.Add(...)](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509certificatecollection.add?view=netframework-4.8#System_Security_Cryptography_X509Certificates_X509CertificateCollection_Add_System_Security_Cryptography_X509Certificates_X509Certificate_))

I'm just concerned that a byte array is way too general, especially when the first thing we do with it is pass it to the `X509Certificate2` constructor.

AdrianJSClark:
This is a good pull request, @ken-wolfe. Thanks for the contribution.

I just had one question for you (see above) before I merge it.

src/dbup-postgresql/PostgresqlConnectionManager.cs
ken-wolfe: The <code>X509CertificateCollection</code> class is intended to be used with the certificate store, which is not the use case that we have. We’re definitely not intending to permanently write the certificate into the store or read it from the store for our use case.

The <code>X509Certficate2</code> constructor does validation of the bye array that is passed in. So I think we should be safe in that respect. It is up to the caller to pass in a valid byte array though. I liked the byte array method because is enables people who have different use cases (file on disk in the clear and file on disk encrypted). If there was demand for an API to take the filename, that could easily be added and would not conflict with this change (though at that point I would also look at upgrading to a later version of npgsql which supports taking the filename in the connection string or using environment variables). Upgrading npgsql does not solve my particular use case, and the upgrade is across a major version with breaking changes, so it should be well understood before doing it.

Here is our use case: We store the certificate encrypted on disk and it gets packaged with our binaries. A docker container with our binaries is made and deployed with kubernetes to run DbUp. We don’t want or need to write the unencrypted file to disk to read it back as it complicates the management of the containers.

sungam3r: > If there was demand for an API to take the filename, that could easily be added and would not conflict with this change

So why not just take a parameter of type `X509Certficate2` and let the caller decide how to create it?

ken-wolfe: Sounds good. Let's just pass in the X509Certificate2. That solves the concern that Adrian brought up and seems plenty flexible. I'll post something for that shortly.

ken-wolfe: I addressed the feedback and re-ran my local tests. Everything seems to be working as expected.

src/dbup-postgresql/PostgresqlExtensions.cs
```diff
@@ -126,6 +157,11 @@ public static void PostgresqlDatabase(this SupportedDatabasesForEnsureDatabase s
 
         using (var connection = new NpgsqlConnection(masterConnectionStringBuilder.ConnectionString))
         {
+            if (certificate!= null)
```
sungam3r:
```suggestion
            if (certificate != null)
```

ken-wolfe:
I approved the last Ivan and the build failed because it couldn't find cake.exe. Is there a way to re-kick the build and checks? I didn't find one by looking around...

ken-wolfe:
@AdrianJSClark or @sungam3r Is there anything else left to do before you can approve this change? I believe I've addressed all the feedback and all of the checks are passing. Thanks again for your prompt consideration.

src/dbup-postgresql/PostgresqlExtensions.cs
```diff
@@ -87,6 +101,18 @@ public static void PostgresqlDatabase(this SupportedDatabasesForEnsureDatabase s
         PostgresqlDatabase(supported, connectionString, new ConsoleUpgradeLog());
```
sungam3r:
```suggestion
        PostgresqlDatabase(supported, connectionString, null);
```

AdrianJSClark:
Was only waiting on having enough time to open GitHub during end-of-year work madness / Christmas Party season.

Thanks for the contribution! I don't have a time frame on this hitting NuGet in a released fashion as I'm not across the current build + release arrangements. @droyad might be able to jump in and answer that though.

PR #476, opened 2019-12-30
Fixed duplicate key exception when retrying PerformUpgrade

MvRens:
During service startup, if the SQL server is unreachable we retry DbUp after a few seconds until it succeeds. We noticed that after a few attempts we got a key violation exception instead of the connection timeout, which could be traced back to the PerformUpgrade exception handler which adds the script name to the Exception data.

If we use "pooling=false" in the connection string this issue does not occur. It looks like the Exception is sometimes re-used by the SqlConnection. This fix works around it by checking the Data dictionary, so we at least get the original exception back.

### DIFF
--- a/src/dbup-core/Engine/UpgradeEngine.cs
+++ b/src/dbup-core/Engine/UpgradeEngine.cs
@@ -95,7 +95,14 @@ namespace DbUp.Engine
             }
             catch (Exception ex)
             {
-                ex.Data.Add("Error occurred in script: ", executedScriptName);
+                const string scriptNameKey = "Error occurred in script: ";
+
+                // Exception can be re-used when using connection pooling
+                if (ex.Data.Contains(scriptNameKey))
+                    ex.Data[scriptNameKey] = executedScriptName;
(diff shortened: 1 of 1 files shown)


src/dbup-core/Engine/UpgradeEngine.cs
```diff
@@ -95,7 +95,14 @@ public DatabaseUpgradeResult PerformUpgrade()
             }
             catch (Exception ex)
             {
-                ex.Data.Add("Error occurred in script: ", executedScriptName);
+                const string scriptNameKey = "Error occurred in script: ";
+
+                // Exception can be re-used when using connection pooling
+                if (ex.Data.Contains(scriptNameKey))
```
sungam3r: You can always use `ex.Data[scriptNameKey] = executedScriptName;`, no need to use if/else

MvRens: Today I learned... thanks for the review! Added a new commit with the updated simpler change.

PR #477, opened 2020-01-16
Added a new transaction strategy that does rollback at the end of transaction

einsteine89:
Wery usefull when
-  you want to test if the scripts won't fail
-  to integrate in a CI pipeline. Fail the build if any of the scripts fail to execute

It is like running in the command line 'dotnet new console --dry-run' where you see what files are going to be created by they wont be created

### DIFF
--- a/src/dbup-core/Engine/Transactions/SingleTransactionWithRollbackStrategy.cs
+++ b/src/dbup-core/Engine/Transactions/SingleTransactionWithRollbackStrategy.cs
@@ -1,0 +1,87 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Data;
+using DbUp.Engine.Output;
+
+namespace DbUp.Engine.Transactions
+{
+    class SingleTransactionWithRollbackStrategy : ITransactionStrategy
+    {
+        IDbConnection connection;
+        IDbTransaction transaction;
+        bool errorOccured;
+        IUpgradeLog log;
+        SqlScript[] executedScriptsListBeforeExecution;
+        List<SqlScript> executedScriptsCollection;
+
+        public void Execute(Action<Func<IDbCommand>> action)
+        {
+            if (errorOccured)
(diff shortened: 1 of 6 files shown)


AdrianJSClark:
Thanks for the contribution, @einsteine89!

One small suggestion, instead of `SingleTransactionWithRollback` would you mind calling it `SingleTransactionAlwaysRollback` ? I find that `SingleTransactionAlwaysRollback` describes better to me that it will rollback the transaction under all circumstances.

The build has failed because you need to alter the approval tests for the public API. If you execute the unit tests you should be prompted to update the approved text file. That will need to be added to your PR to merge it.

AdrianJSClark:
Open for discussion on the name, BTW.

src/dbup-core/Engine/Transactions/SingleTransactionWithRollbackStrategy.cs
```diff
@@ -0,0 +1,87 @@
...
+            else
+            {
+                log.WriteWarning("Error occured when executing scripts, transaction will be rolled back");
+            }
+
+            //Restore the executed scripts collection
+            executedScriptsCollection.Clear();
+            executedScriptsCollection.AddRange(executedScriptsListBeforeExecution);
+
+            transaction.Dispose();
```
sungam3r: Explicitly call `Rollback` ?

einsteine89: I just forgot about that. It is worth adding Rollback, even if at Dispose it will be rollbacked if Commit was not explicitly called. Just added that.

einsteine89:
Hi! What are the chances to have this functionality available in a new release?

AdrianJSClark:
I don't know what the status is with releases, @einsteine89. Do you have any idea about the roadmap you can comment on quickly here, @droyad?

Issue #325, opened 2018-04-04
Output SQL that is being run


phillijw:
Is there a way to output the finalized SQL that gets run? I did not see any way to log this. Debugging issues is impossible because it seems the executors are rewriting sql for various reasons but I can't determine what the result of that rewrite is without outputting it somehow.

JustinPealing:
You can use the `Scripts` property on `DatabaseUpgradeResult`, like so:

    foreach (var script in result.Scripts)
    {
        Console.WriteLine("==============================");
        Console.WriteLine($"{script.Name}");
        Console.WriteLine("==============================");
        Console.WriteLine(script.Contents);
        Console.WriteLine();
    }

There is also an extension method `WriteExecutedScriptsToOctopusTaskSummary` that does something similar specifically for use in Octopus deploy:

    result.WriteExecutedScriptsToOctopusTaskSummary();

Note however that if a script encounters an error, it won't be included in that collection. To work out what script will have caused the error you can use `GetScriptsToExecute` on the engine, and then compare to the scripts in the result:

    var scriptsToExecute = engine.GetScriptsToExecute();
    var result = engine.PerformUpgrade();
    
    // If an error happens, then the script that errors does not appear in result.Scripts. Check the list of scripts
    // that we know should have been executed to identify and log the first script that didn't run.
    var notRun = scriptsToExecute.FirstOrDefault(s => result.Scripts.All(x => x.Name != s.Name));
    if (notRun != null)
    {
        Console.WriteLine("==============================");
        Console.WriteLine($"FAILED - {notRun.Name}");
        Console.WriteLine("==============================");
        Console.WriteLine(notRun.Contents);
        Console.WriteLine();
    }

JustinPealing:
It would be handy to have some sort of property on the result that made it easier to get the script that was running when an exception was thrown. I could submit a PR?

mjauernig:
Seems to be done already. If still needed, please reopen.

PR #478, opened 2020-01-20
Write failed script to DatabaseUpgradeResult

cjconrad2:
Took a shot at #325 by adding an ErrorScript property to DatabaseUpgradeResult. 

Fails the public API change test, but judging by the suggested fix on the issue, I think that is to be expected. Not sure what your process is for updating those tests if this change is accepted, but happy to do it with instruction.


sungam3r:
Just change approved txt file.

PR #497, opened 2020-04-24
Oracle adapter does not correctly handle custom statement

aszego:
This PR fixes an issue with the Oracle adapter when using custom statement separator via the `DELIMITER` keyword.
The issue was that this keyword was ignored if it was right at the beginning of a script.
Added unit test as well to avoid a regression in the future.


sungam3r:
> The issue was that this keyword was ignored if it was right at the beginning of a script.
Added unit test as well to avoid a regression in the future.

Could you please add one more test to verify delimiter keyword not in the beginning of a script?

droyad:
Thanks @aszego, I covered this in #509

PR #507, opened 2020-06-04
Oracle commands should not be split on ; as it will screw up scripts.…

biztechprogramming:
…  They should be split on forward slash.


sungam3r:
Duplicate of #335 ?

droyad:
Closed as duplicate

PR #511, opened 2020-06-26
Fixed NRE on PerformUpgrade error handling

sombre-hombre:
When connection error has occurred and no scripts executed, there is NullReferenceException on error handling block.


AdrianJSClark:
Nice catch, thanks for the contribution!

PR #515, opened 2020-07-11
Add option not to prefix scripts with subdir names

jez9999:
I want to organize my scripts into subdirectories but still sort just on filename, so I've added this option not to include subdir name in script names with loading from subdirectories.

### DIFF
--- a/src/dbup-core/ScriptProviders/FileSystemScriptProvider.cs
+++ b/src/dbup-core/ScriptProviders/FileSystemScriptProvider.cs
@@ -79,12 +79,14 @@ namespace DbUp.ScriptProviders
                         new DirectoryInfo(directoryPath).GetFiles(scriptExtension, ShouldSearchSubDirectories())
                     );
                 }
-                foreach (var fileInfo in files)
-                {
-                    if (files.Count(f => f.Name == fileInfo.Name) > 1)
-                    {
-                        throw new Exception($"Duplicate filename: {fileInfo.Name}");
+                var dupeFiles = files.GroupBy(f => f.Name).Where(grp => grp.Count() > 1).SelectMany(f => f.Select(fi => fi.FullName)).ToList();
+                if (dupeFiles.Count > 0) {
+                    var sbError = new StringBuilder();
+                    sbError.AppendLine("Duplicate filenames:");
+                    foreach (var file in dupeFiles) {
+                        sbError.AppendLine($"- {file}");
                     }
+                    throw new Exception(sbError.ToString());
                 }
                 if (filter != null)
                 {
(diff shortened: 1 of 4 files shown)


src/dbup-core/ScriptProviders/FileSystemScriptProvider.cs
```diff
@@ -55,18 +55,45 @@ public FileSystemScriptProvider(string directoryPath, FileSystemScriptOptions op
...
+                {
+                    files.AddRange(
+                        new DirectoryInfo(directoryPath).GetFiles(scriptExtension, ShouldSearchSubDirectories())
+                    );
+                }
+                foreach (var fileInfo in files)
+                {
+                    if (files.Count(f => f.Name == fileInfo.Name) > 1)
+                    {
+                        throw new Exception($"Duplicate filename: {fileInfo.Name}");
```
AdrianJSClark, line 86: This exception should have the full path of the duplicate files.

Can I suggest using a LINQ `.GroupBy(...)` to detect multiple names combined with a custom exception that could list all the duplicate file paths as well as the name, as done here? That way a consumer of the API can get information about which files are being included. It may not be obvious just from the individual file name.

src/dbup-tests/ApprovalFiles/dbup-core.approved.cs
```diff
@@ -409,6 +409,7 @@ public class FileSystemScriptOptions
         public string[] Extensions { get; set; }
         public System.Func<string, bool> Filter { get; set; }
         public bool IncludeSubDirectories { get; set; }
+        public bool PrefixWithSubDirectoryNames { get; set; }
```
AdrianJSClark, line 412: While in the context of this PR I understand what this option switch does, I think looking at the `FileSystemScriptOptions` in isolation the name would be unclear.

Perhaps `UseFullPathAsScriptName` is better? Happy to take suggestions.

Can I please ask that you add a documentation comment (i.e. `/// <summary>...</summary>` ) with a quick explanation on what this option does. I know the other options do not have that, however it would be great if new features could include it. Most IDEs will expose that information as a tool tip or similar making everything more approachable.

jez9999: Trouble is it's not the full path; it's a relative path like "sd1.sd2.filename"

AdrianJSClark: Oh, OK. That makes that a bad name then!

Perhaps flip it around and make it `UseOnlyFileNameForScriptName` with a default of `false` for the current functionality and `true` for your new functionality.

What do you think?

jez9999: Yeah, sounds good.

jez9999:
@AdrianJSClark Have addressed your feedback, hopefully now this PR is good!

AdrianJSClark:
Looks good! Thanks for the changes.

PR #518, opened 2020-08-06
Remove line break on script output.

pitermarx:
Any special reason for this?
I do not think it is necessary. The PRINT statements in SQL already include the newline.
Feel free to dismiss this though.


AdrianJSClark:
Probably included out of habit. I don't see a reason to have the extra line ending.

AdrianJSClark:
Thanks for the contribution, @pitermarx!

pitermarx:
BTW, thanks for the project. Any idea on the timeframe for releasing 5.0?

AdrianJSClark:
Not sure at all, unfortunately. Will need to chat with @droyad and see how things look.

Issue #479, opened 2020-01-21
[DbUp.Oracle] Script fail when comment at end


Orkad:
Don't know if its only oracle related but in fact, if you try to run the following scripts :
`create table FOO (myid INT NOT NULL); -- this comment will fail the script`
`create table FOO (myid INT NOT NULL); /* this comment will fail the script too */`

the scripts will fail according to the oracle exception ORA-00900 invalid SQL statement. 

DbUp consider the comment as SQL instruction in each cases.

sungam3r:
What version do you use?

Orkad:
4.3.0 (current latest)

Orkad:
![image](https://user-images.githubusercontent.com/921052/72841026-cddfa080-3c95-11ea-8d2d-64629b6a5bff.png)

Actualy this simple test fail. Do you want a PR of that ?

sungam3r:
Interesting. Perhaps this is due to \r\n absence at the end of script? It looks like a bug.

Orkad:
Same problem with `Environnement.NewLine` at the end of script. I think the sql parser just split our script relative to the dialect delimiters. It will be interresting to remove comment from script before execution, or maybe detect if parsed text is a command.

PR #520, opened 2020-08-14
SqlCommandReader add ability to ignore comments

Orkad:
I found a way to resolve the #479 bug. 
Its a tiny upgrade introducted with an optionnal parameter "ignoreComments" in SqlCommandReader.
The default is setted to false to keep existing process.

I spread the parameter to oracle support but i can go deeper !


AdrianJSClark:
This looks like a good work-around @Orkad.

The check that is likely failing is the test which checks public API changes. You should see that fail locally as well. You'll need to commit the appropriate "*.approved.txt" file updated with your additional parameter to get the test to pass.

Orkad:
Sorry for the time it takes. Do you want me to do the same thing for MySqlCommandReader ?

PR #528, opened 2020-09-23
Pass script options to embeddedScriptProvider

fire-birdie:
The `EmbeddedScriptAndCodeProvider` was creating its internal `embeddedScriptProvider` with the default `SqlScriptOptions` instead of the instance passed by the caller. As a result, custom values for `ScriptType` and `RunGroupOrder` were being applied to code-based scripts, but not embedded script files.

### DIFF
--- a/src/dbup-core/ScriptProviders/EmbeddedScriptAndCodeProvider.cs
+++ b/src/dbup-core/ScriptProviders/EmbeddedScriptAndCodeProvider.cs
@@ -60,7 +60,7 @@ namespace DbUp.ScriptProviders
             this.assembly = assembly;
             this.filter = codeScriptFilter ?? filter;
             this.sqlScriptOptions = sqlScriptOptions;
-            embeddedScriptProvider = new EmbeddedScriptProvider(assembly, filter);
+            embeddedScriptProvider = new EmbeddedScriptProvider(assembly, filter, DbUpDefaults.DefaultEncoding, sqlScriptOptions);
         }
 
         IEnumerable<SqlScript> ScriptsFromScriptClasses(IConnectionManager connectionManager)


src/dbup-core/ScriptProviders/EmbeddedScriptAndCodeProvider.cs
```diff
@@ -60,7 +60,7 @@ public EmbeddedScriptAndCodeProvider(Assembly assembly, Func<string, bool> filte
             this.assembly = assembly;
             this.filter = codeScriptFilter ?? filter;
             this.sqlScriptOptions = sqlScriptOptions;
-            embeddedScriptProvider = new EmbeddedScriptProvider(assembly, filter);
+            embeddedScriptProvider = new EmbeddedScriptProvider(assembly, filter, DbUpDefaults.DefaultEncoding, sqlScriptOptions);
```
evansmith1377, line 63: Not related to #544, but should we pass `this.filter` to the new `EmbeddedScriptProvider` since the XML comment says "If null, filter is used"?

evansmith1377:
@AdrianJSClark would it be possible to have someone look at this pull request and advise if it's a suitable change? It looks good to me, but I've only been in the source for a day or so.

(btw, I certainly appreciate the effort you all have put into DbUp; it is a great solution!)

src/dbup-core/ScriptProviders/EmbeddedScriptAndCodeProvider.cs
sungam3r: Anyway two filters look unusual.

sungam3r (approved):

Issue #524, opened 2020-09-01
Shouldn't this be disposed?


pitermarx:
https://github.com/DbUp/DbUp/blob/a0a9c6ddbe2c107d13f7e66e319248ede0709ce6/src/dbup-core/Engine/Transactions/DatabaseConnectionManager.cs#L71

PeteBoyRocket:
I agree, I think it should be disposed too. Like this? https://github.com/DbUp/DbUp/pull/529

sungam3r:
Already desposed in `master`.

PR #529, opened 2020-09-26
Dispose of connection in TryConnect method

PeteBoyRocket:
Fix for issue: https://github.com/DbUp/DbUp/issues/524


AdrianJSClark:
Thanks for fixing this up.

PR #531, opened 2020-09-29
Update variable-substitution.md : add extra information on reserved variable name 'schema'.

jonathanantoine:
Add extra information on reserved variable name.


AdrianJSClark:
Thanks for the contribution! Can imagine that's a trap people may fall into.

Probably a candidate for cleaning up one day as that variable substitution probably should happen alongside the others.

jonathanantoine:
Ou team Lost one day figuring out what was happening 😁

PR #550, opened 2021-01-14
Add MarkAsExecuted predicate overload

LeeReitz:
Added predicate overload to UpgradeEngine.MarkAsExecuted(). My use case for this change is to mark a subset of scripts as executed based on environment based on the script name prior to performing upgrade. IE in prod, don't apply development and staging scripts. Used predicate to allow many use cases where a subset is needed.


sungam3r:
@LeeReitz  Please rebase onto `master` if you are still interested.

PR #552, opened 2021-01-16
Fix potential infinite loop in SkipWhitespace

### LINKED ISSUE #551: Potential infinite loop in MySqlCommandReader SkipWhitespace
No check is made for a failed read which can result in an infinite loop.

napalu:
Check for FailedResult after Read()
closes #551


AdrianJSClark:
Thanks for fixing this up. Extra thanks for including a test, it is appreciated.

PR #559, opened 2021-02-12
Enable Alternate Azure SQL Instances

### LINKED ISSUE #555: Add Azure Gov as option for Azure Sql Integrated Security.
We are using Azure gov and I noticed the database token access url is hardcoded in the SqlConnectionManger.  The Azure gov url is https://database.usgovcloudapi.net.  

So either add another parameter for gov, overloaded method for gov, or a way to pass our own token or make the url configurable.  Any options would solve the problem. 

image

AdrianJSClark:
Certain alternate Azure SQL Instances, such as those specifically for Government use, need additional configuration to be passed through to the Azure Service Token Provider in order to properly connect.

This change implements additional extension methods which accept the specialised parameters to allow the connection to be properly set up.

Fixes #555 "Add Azure Gov as option for Azure Sql Integrated Security"

### DIFF
--- a/src/dbup-sqlserver/AzureSqlConnectionManager.cs
+++ b/src/dbup-sqlserver/AzureSqlConnectionManager.cs
@@ -1,0 +1,48 @@
+﻿using System.Collections.Generic;
+using System.Data.SqlClient;
+using DbUp.Engine.Transactions;
+using DbUp.Support;
+
+#if SUPPORTS_AZURE_AD
+using Microsoft.Azure.Services.AppAuthentication;
+
+namespace DbUp.SqlServer
+{
+    /// <summary>Manages an Azure Sql Server database connection.</summary>
+    public class AzureSqlConnectionManager : DatabaseConnectionManager
+    {
+        public AzureSqlConnectionManager(string connectionString)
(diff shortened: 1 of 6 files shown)


src/dbup-sqlserver/AzureSqlConnectionManager.cs
```diff
@@ -0,0 +1,48 @@
...
+        public AzureSqlConnectionManager(string connectionString)
+            : this(connectionString, "https://database.windows.net/", null)
+        { }
+
+        public AzureSqlConnectionManager(string connectionString, string resource)
+            : this(connectionString, resource, null)
+        { }
+
+        public AzureSqlConnectionManager(string connectionString, string resource, string tenantId, string azureAdInstance = "https://login.microsoftonline.com/")
+            : base(new DelegateConnectionFactory((log, dbManager) =>
```
chrismcclure, line 23: Yep this is what I was thinking of.

chrismcclure (approved):
This should solve the problem that I was having.

Boruc04:
@AdrianJSClark any chance that this will be released via nuget any time soon?

PR #561, opened 2021-02-19
Make UpgradeEngine and UpgradeEngineBuilder virtual to allow custom b…

nmangue:
For some special use cases, it can be useful to extend `UpgradeEngine` to add some custom behavior. In my case, I would like to execute some scripts after `UpgradeEngine.PerformUpgrade` for multiple applications I maintain.
To do so, `UpgradeEngine` and `UpgradeEngineBuilder` methods need to be virtual.


shep1987:
This would be really useful to me, can I grab a pre-release version from anywhere?

PR #563, opened 2021-02-20
Add SQLite journal extension

SteveDesmond-ca:
This PR adds an extension/helper method so SQLite databases can use the same fluid builder syntax as SQL Server, PostgreSQL, and Redshift.

It also fixes the comment from the PostgreSQL version, which I'm assuming was copied and pasted from SQL Server (as I did for the SQLite one :sweat_smile:)

I noticed the existing methods had no test coverage, so lmk if you'd like that added!

### DIFF
--- a/src/dbup-postgresql/PostgresqlExtensions.cs
+++ b/src/dbup-postgresql/PostgresqlExtensions.cs
@@ -207,7 +207,7 @@ public static class PostgresqlExtensions
     }
 
     /// <summary>
-    /// Tracks the list of executed scripts in a SQL Server table.
+    /// Tracks the list of executed scripts in a PostgreSQL table.
     /// </summary>
     /// <param name="builder">The builder.</param>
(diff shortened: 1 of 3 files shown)


stevedesmond-ca:
Also can't quite figure out why the approval tests are failing, but it looks like they're also failing on `master`

AdrianJSClark:
Thanks for this contribution, @stevedesmond-ca!

Apologies, it was some hasty merging from me which broke the build due to the tests. I've posted a PR (#564) which should hopefully resolve the issue. I'm hoping for someone to double check my work before merging, then we can apply those changes to this branch or rebase it and hopefully resolve the build problems here too.

SteveDesmond-ca:
It looks like my fork got disconnected from this PR, so I'll create a new one with these same changes rebased off `HEAD` now that #564 is merged.

PR #564, opened 2021-02-21
Fix Public API Ordering For Acceptance Tests

AdrianJSClark:
Sometimes between builds on my client PC and the build server the order that some of the API definitions are generated varies. This change does the following things to attempt to make the sorting of API components more explicit and repeatable:

 - Explicitly uses the "InvariantCulture" ordering for string values
 - Adds method sorting by return type name, then number of parameters

This PR also includes the updated acceptance files for the public API with the above changes made.

These changes are designed to fix the currently-broken build.

### DIFF
--- a/src/dbup-tests/ApiTests.cs
+++ b/src/dbup-tests/ApiTests.cs
@@ -59,7 +59,7 @@ namespace DbUp.Tests
                              where t.IsPublic
                              group t by t.Namespace;
 
-            foreach (var ns in namespaces.OrderBy(n => n.Key))
+            foreach (var ns in namespaces.OrderBy(n => n.Key, StringComparer.InvariantCulture))
             {
                 if (ns.Key == null)
                 {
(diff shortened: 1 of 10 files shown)


sungam3r:
Why not just use [PublicApiGenerator](https://www.nuget.org/packages/PublicApiGenerator/) ?

Example - https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL.ApiTests/ApiApprovalTests.cs

AdrianJSClark:
@sungam3r This code existed within DbUp for a very long time. I don't want to make big changes when it isn't necessary.

sungam3r:
Whether it is necessary or not is up to you. I just saw that you are editing code that has long been in PublicApiGenerator and is well tested. It all comes down to only the desire of the maintainers of this repo to continue to support the manually written code for public API generation which is obviously not the main task for DbUP.

In addition, some parts of the generated code could be simplified, for example all lines like
`public System.Nullable<int> ExecutionTimeoutSeconds { get; set; }`
could be simplified to
`public int? ExecutionTimeoutSeconds { get; set; }`

PublicApiGenerator handles this. Also PublicApiGenerator knows how to work with NRT. If DbUP one day starts to use NRT, then I very much doubt that you will want to manually write the code to support it. This is a non-trivial task.

AdrianJSClark:
This code has worked for a long time. Certainly if it had some big deficiencies we could consider other options, this was merely a few minor fixes.

If you feel there's a strong case for the upgrade we'd be happy to accept a PR.

sungam3r:
No strong case.

Issue #558, opened 2021-02-12
Question: plan to move off dotnetframework?


Justin-DynamicD:
Are there any plans to get off of dotnet framework so that this tool can be run from multiple operating systems?  I have a lot of clients now that do all their builds and work from linux-based agents, so being able to bundle this tool into those build agents would be a big win.

AdrianJSClark:
I don't think it is realistic to move completely away from .NET Framework support. There are lots of projects out there that are still using it very successfully with no plans to migrate away so DbUp should keep supporting them. This doesn't mean that the code can't be written in a way to support your goals.

There was discussion about moving to newer versions to allow use of newer features and reduce complexity in #432. I think if the repository split discussed in #431 was implemented we could take that opportunity of the major version bump (which would be required) to not only get rid of some of the very old platforms but also to make moves towards newer, cross-platform support.

For example, as part of the major release the `DbUp.SqlServer` package could change it's references to use the [Microsoft.Data.Client](https://www.nuget.org/packages/Microsoft.Data.SqlClient/) package instead of using `System.Data.Client`. This would break the Windows requirement for that particular database technology.

Justin-DynamicD:
I read about the split and nearly commented in there on this issue, but felt it wasn't appropriate so made this ask instead.

Thanks for the viewpoint on development.  I would love to see a push for a newer standard, as at the end of the day most consumers of the tool are just downloading and executing a binary ... so they really don't care much about how it was built/compiled. I have clients right now who use MSSQL out of familiarity, but their app stack is all go.  They couldn't care less about what's under the hood ... just that dbup lets them push sql scripts in a structured manner. *shrug*.

stevenwilliamsmis:
@AdrianJSClark I submitted a pull request yesterday before I saw this post #565 to add support of the Microsoft.Data.SqlClient package using compile flags to maintain backwards compatibility for Net 3.5 and Standard 1.3. 
 What do you mean when you stated above that updating would break the windows requirement?

AdrianJSClark:
I meant that `Microsoft.Data.SqlClient` is designed to be cross-platform, however `System.Data.SqlClient` is not. So the dependency on Windows is broken when we upgrade.

I did see #565 come through, thanks for posting that.

AdrianJSClark:
There is a `net6.0` target available which references `Microsoft.Data.SqlClient` as of #652.

PR #565, opened 2021-02-22
Add support for Microsoft.Data.SqlClient

stevenwilliamsmis:
System.Data.SqlClient is now moved to Microsoft.Data.SqlClient and while System.Data.SqlClient still works all new development is in the new library as per the GitHub page [https://github.com/dotnet/SqlClient](https://github.com/dotnet/SqlClient). The upgrade is simple as it is just namespace changes however the new library only supports .NetStandard 2.0 and .Net Framework 4.6. I have added a new compile flag to allow for backwards compatibility of .Net 3.5 and .NetStandard 1.3 as they will continue to use their current methods of accessing SqlClient.

The feature I'm after is provided in the latest release in Microsoft.Data.SqlClient 2.1 called Azure Active Directory Managed Identity authentication it just requires a property in the connection string to allow managed Identity access to Azure sql server [https://github.com/dotnet/SqlClient/blob/master/release-notes/2.1/2.1.0.md](https://github.com/dotnet/SqlClient/blob/master/release-notes/2.1/2.1.0.md)

### DIFF
--- a/src/dbup-sqlserver/dbup-sqlserver.csproj
+++ b/src/dbup-sqlserver/dbup-sqlserver.csproj
@@ -24,21 +24,22 @@
     <ProjectReference Include="..\dbup-core\dbup-core.csproj" />
   </ItemGroup>
 
-  <ItemGroup Condition=" '$(TargetFramework)' == 'net35' Or '$(TargetFramework)' == 'net46' ">
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net35' ">
     <Reference Include="System.Data" />
     <Reference Include="System" />
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'net46' ">
+    <PackageReference Include="Microsoft.Data.SqlClient" Version="2.1.1" />
     <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.3.1" />
   </ItemGroup>
-
+  
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.3' ">
     <PackageReference Include="System.Data.SqlClient" Version="4.4.0" />
   </ItemGroup>
-
+  
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
-    <PackageReference Include="System.Data.SqlClient" Version="4.6.0" />
+    <PackageReference Include="Microsoft.Data.SqlClient" Version="2.1.1" />
     <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.3.1" />
   </ItemGroup>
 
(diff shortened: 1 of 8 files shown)


src/dbup-sqlserver/dbup-sqlserver.csproj
```diff
@@ -30,15 +30,16 @@
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'net46' ">
+    <PackageReference Include="Microsoft.Data.SqlClient" Version="2.1.1" />
     <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.3.1" />
   </ItemGroup>
-
+
```
sungam3r, line 36:
```suggestion
```

src/dbup-sqlserver/dbup-sqlserver.csproj
```diff
@@ -30,15 +30,16 @@
...
+    <PackageReference Include="Microsoft.Data.SqlClient" Version="2.1.1" />
     <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.3.1" />
   </ItemGroup>
-
+  
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.3' ">
     <PackageReference Include="System.Data.SqlClient" Version="4.4.0" />
   </ItemGroup>
-
+
```
sungam3r, line 40:
```suggestion

```

AdrianJSClark:
Thanks for this submission!

You have some issues with our checks to ensure that the public API of DbUp remains unchanged unless we intend it to change. You'll need to run the tests locally where it should prompt you with a file diff between the "received" and "approved" versions. If you update the "approved" file with the public API changes you intend to make the tests will pass.

I haven't had time to look at the PR more deeply. Just on the surface though I'm a little concerned that people may be confused about which version of the SQL Client libraries they are getting if we simply swap things out with a compile flag. This was the reason that I mentioned in #558 that this change would be a major version upgrade.

Give me some time to look into the detail. Not sure if @droyad has any input or suggestions he would like to share. At the very least I suggest we'd need to commit to version 5.0 of DbUp to merge this change.

stevenwilliamsmis:
@AdrianJSClark thanks for the feedback.  I thought it was due to the test breaking from another pull request that was accepted into main and I overlooked the approve files. I see the missing files, will commit.

stevenwilliamsmis:
So I struggled a little with the approval files when I run the test locally I do not get the received files that you mentioned.  The tests for sql server were failing with the following error: System.Exception: Could not find a diff program to use.  The strange thing is that all of the other api test, with the exception of the upgrade engine api test which was failing in main, are running so if it was a tooling issue I would have thought that they all would fail.
The other thing I found was that dbup exposes the SqlConnectionStringBuilder class in the TemporarySqlDatabase class as a constructor parameter.  The fix for the SQL server api test was to set the fully qualified name to Microsoft.Data.SqlClient.SqlConnectionStringBuilder but this is not fully correct as the test is only covering .net46 and .netstandard 2.0 not sure how to proceed with the test modification. 
As far as your concern that people may be confused about which version of the SQL Client libraries they are getting does it really matter? Except for that one constructor that exposes the SqlConnectionStringBuilder, which I question why that is exposed to begin with,  the user shouldn't even need know what is being used behind the scenes. This change wouldn't force a user to update any other code outside of that parameter as they could have both libraries installed to maintain their code base outside of dbup.  Thoughts?

Hooch180:
When will the support for Microsoft.Data.SqlClient be added?

uladz-zubrycki:
Are there any plans to get this one merged?

AdrianJSClark:
A version of this was merged in #652. You will need to target `net6.0` to pick up `Microsoft.Data.SqlClient` support right now, which seemed the most-compatible way of getting this out. This will be released soon as a version 5 of `dbup-sqlserver`.

Please open an issue and let us know if this causes an issue for you and you really need to be able to reference `Microsoft.Data.SqlClient` from .NET Framework.

I'll close this PR now as the aim of it has been addressed.

PR #567, opened 2021-03-11
Support spaces between connection string separators [;=]

mauricio-morales:
I recently ran into a _bug_ where DbUp wasn't creating the `schemaversions` table on my schema cause it was finding another `schemaversions` in another schema. 

Turns out my connection string looked something like this:

```
data source=dbserver.com; database=myschema
```

That ` ` (space) after `;` was causing the `TableJournal` to not recognize the explicit schema. This PR fixes that bug and also accounts for potential spaces on the value-side of the database (i.e. `; database = myschema`).

### DIFF
--- a/src/dbup-mysql/MySqlExtensions.cs
+++ b/src/dbup-mysql/MySqlExtensions.cs
@@ -24,9 +24,9 @@ public static class MySqlExtensions
     /// </returns>
     public static UpgradeEngineBuilder MySqlDatabase(this SupportedDatabases supported, string connectionString)
     {
-        foreach (var pair in connectionString.Split(';').Select(s => s.Split('=')).Where(pair => pair.Length == 2).Where(pair => pair[0].ToLower() == "database"))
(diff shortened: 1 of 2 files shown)


mjauernig:
The CI fails because of 
```
xUnit.net 00:00:05.64]     DbUp.Tests.Providers.MySql.DatabaseSupportTests.VerifyBasicSupport [FAIL]
[xUnit.net 00:00:05.64]       Assent.AssentApprovedFileNotFoundException : The assent file '\_\src\dbup-tests\Support\MySql\ApprovalFiles\DatabaseSupportTests.VerifyBasicSupport.approved.txt' was not found
```
image

On the repository the [file ](https://github.com/DbUp/DbUp/blob/master/src/dbup-tests/Support/MySql/ApprovalFiles/DatabaseSupportTests.VerifyBasicSupport.approved.txt ) exists. 

I will take a deeper look onto die CI the next days.

PR #575, opened 2021-05-19
Update Npgsql 3.2.6 -> 3.2.7 to enable SCRAM-SHA-256 auth support

Belorus:
SCRAM-SHA-256 is a recommended auth mode since postgresql 10 that replaces md5.

See https://github.com/npgsql/npgsql/issues/1530

### DIFF
--- a/src/dbup-postgresql/dbup-postgresql.csproj
+++ b/src/dbup-postgresql/dbup-postgresql.csproj
@@ -33,7 +33,7 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' != 'net35'">
-    <PackageReference Include="Npgsql" Version="3.2.6" />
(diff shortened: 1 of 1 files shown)


TroyWitthoeft:
Is there something blocking this?  
Can we upgrade DbUp from the 2017 version of npgsql 3.2.6? 
Latest npgsql version is 5.0.7 and it's picked up plenty of security and feature enhancements.  Most notable about those that support "new sql" distributed cloud dbs like spanner and cockroach.

jdiazmartin:
It would be nice to upgrade the npgsql version, the current version has problems with the connection string field: ssl_renegotiation_limit with the error `Npgsql.PostgresException (0x80004005): 08P01: unsupported startup parameter: ssl_renegotiation_limit`.

Belorus:
This was done in 4.6.0

PR #576, opened 2021-06-09
Allow usage of an existing sql connection

ipinak:
Updated `SqlConnectionManager` and `SqlServerExtensions` to allow an existing SqlConnection to be used for migrating.

### DIFF
--- a/src/dbup-sqlserver/SqlConnectionManager.cs
+++ b/src/dbup-sqlserver/SqlConnectionManager.cs
@@ -15,15 +15,7 @@ namespace DbUp.SqlServer
         /// </summary>
         /// <param name="connectionString"></param>
         public SqlConnectionManager(string connectionString)
-             : base(new DelegateConnectionFactory((log, dbManager) =>
-             {
-                 var conn = new SqlConnection(connectionString);
-
-                 if (dbManager.IsScriptOutputLogged)
-                     conn.InfoMessage += (sender, e) => log.WriteInformation($"{{0}}", e.Message);
-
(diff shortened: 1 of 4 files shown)


src/dbup-sqlserver/SqlConnectionManager.cs
```diff
@@ -15,15 +15,17 @@ public class SqlConnectionManager : DatabaseConnectionManager
...
+             : this(new SqlConnection(connectionString))
+        { }
 
-                 if (dbManager.IsScriptOutputLogged)
-                     conn.InfoMessage += (sender, e) => log.WriteInformation($"{{0}}", e.Message);
+        public SqlConnectionManager(SqlConnection connection)
+               : base(new DelegateConnectionFactory((log, dbManager) =>
+               {
+                   if (dbManager.IsScriptOutputLogged)
+                       connection.InfoMessage += (sender, e) => log.WriteInformation($"{{0}}", e.Message);
```
sungam3r, line 25: Pay attention to that `+=`. You intend to pass the connection from the outside (`public static UpgradeEngineBuilder SqlDatabase(this SupportedDatabases supported, SqlConnection connection)`) so every time the handler will be appended here multiplying log count.

sungam3r:
@ipinak  Please rebase onto `master` if you are still interested.

tylerh:
Is this something we could get pushed through?

droyad:
Implemented using https://github.com/DbUp/dbup-sqlserver/pull/31

PR #586, opened 2021-09-22
[jhw] fix handling of # comments in mysql query parsing

jonnyhweiss:
I had an issue trying to create a stored procedure because I had a `# this isn't an important comment` comment in the middle of it, and that broke the parsing of it. I see we handle `--` and `/*` comments, but not `#` comments. 

https://dev.mysql.com/doc/refman/5.6/en/comments.html


sungam3r:
@jonnyhweiss I can review again after feedback if you are still interested.

PR #592, opened 2021-11-23
Add SQLite journal extension

SteveDesmond-ca:
(copied from #563)

This PR adds an extension/helper method so SQLite databases can use the same fluid builder syntax as SQL Server, PostgreSQL, and Redshift.

It also fixes the comment from the PostgreSQL version, which I'm assuming was copied and pasted from SQL Server (as I did for the SQLite one :sweat_smile:)

I noticed the existing methods had no test coverage, so lmk if you'd like that added!

### DIFF
--- a/src/dbup-sqlite/SqliteExtensions.cs
+++ b/src/dbup-sqlite/SqliteExtensions.cs
@@ -49,4 +49,15 @@ public static class SQLiteExtensions
         builder.WithPreprocessor(new SQLitePreprocessor());
         return builder;
     }
+    
+    /// <summary>
+    /// Tracks the list of executed scripts in a custom SQLite table.
+    /// </summary>
+    /// <param name="table">The name of the table used to store the list of executed scripts.</param>
+    /// <returns>The <see cref="UpgradeEngineBuilder"/> used to set the journal table name.</returns>
+    public static UpgradeEngineBuilder JournalToSQLiteTable(this UpgradeEngineBuilder builder, string table)
+    {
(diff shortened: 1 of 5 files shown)


sungam3r:
@SteveDesmond-ca  Please rebase onto `master` if you are still interested.

SteveDesmond-ca:
@sungam3r thanks for checking in on this, I had completely forgotten about it -- branch has been rebased, lmk if you need anything else!

src/dbup-sqlite/SqliteExtensions.cs
```diff
@@ -49,4 +49,17 @@ public static UpgradeEngineBuilder SQLiteDatabase(this SupportedDatabases suppor
         builder.WithPreprocessor(new SQLitePreprocessor());
         return builder;
     }
-}
\ No newline at end of file
+    
+    /// <summary>
+    /// Tracks the list of executed scripts in a SQLite table.
+    /// </summary>
+    /// <param name="builder">The builder.</param>
```
mjauernig, line 56: Because it is an extension we can remove the param.

src/dbup-sqlite/SqliteExtensions.cs
```diff
@@ -49,4 +49,17 @@ public static UpgradeEngineBuilder SQLiteDatabase(this SupportedDatabases suppor
...
+    /// <param name="schema">The schema.</param>
+    /// <param name="table">The table.</param>
+    /// <returns></returns>
```
mjauernig, line 59: Missing return description.

mjauernig (commented):
Did i understand it right? The function is for save the journal into another table then "SchemaVersions"?

mjauernig (approved):

Issue #594, opened 2021-12-09
Security vulnerabilities in old version of System.Net.Http


Havunen:
There are security vulnerabilities in old version of System.Net.Http which seems to be used from Microsoft.Azure.Services.AppAuthentication

<details>
<summary><b>Denial of Service (DoS) [System.Net.Http]</b></summary>

Severity: `high`
CVSS Score: `7.5`
Vulnerability chain:

- `dbup-sqlserver@4.5.0`
- `Microsoft.Azure.Services.AppAuthentication@1.3.1`
- `Microsoft.IdentityModel.Clients.ActiveDirectory@4.3.0`
- `NETStandard.Library@1.6.1`
- `System.Net.Http@4.3.0`

Fixed in: `System.Net.Http 4.1.2,4.3.2`
## Overview

[System.Net.Http](https://www.nuget.org/packages/System.Net.Http/) is an Provides a programming interface for modern HTTP applications, including HTTP client components that allow applications to consume web services over HTTP and HTTP components that can be used by both clients and servers for parsing HTTP headers.

Affected versions of this package are vulnerable to Denial of Service (DoS)
as `ASP.NET Core` fails to properly validate web requests.

**NOTE:** Microsoft has not commented on third-party claims that the issue is that the `TextEncoder.EncodeCore` function in the `System.Text.Encodings.Web` package in ASP.NET Core Mvc before 1.0.4 and 1.1.x before 1.1.3 allows remote attackers to cause a denial of service by leveraging failure to properly calculate the length of 4-byte characters in the Unicode Non-Character range.

## Details
Denial of Service (DoS) describes a family of attacks, all aimed at making a system inaccessible to its intended and legitimate users.

Unlike other vulnerabilities, DoS attacks usually do not aim at breaching security. Rather, they are focused on making websites and services unavailable to genuine users resulting in downtime.

One popular Denial of Service vulnerability is DDoS (a Distributed Denial of Service), an attack that attempts to clog network pipes to the system by generating a large volume of traffic from many machines.

When it comes to open source libraries, DoS vulnerabilities allow attackers to trigger such a crash or crippling of the service by using a flaw either in the application code or from the use of open source libraries.

Two common types of DoS vulnerabilities:

* High CPU/Memory Consumption- An attacker sending crafted requests that could cause the system to take a disproportionate amount of time to process. For example, [commons-fileupload:commons-fileupload](SNYK-JAVA-COMMONSFILEUPLOAD-30082).

[...]

PR #595, opened 2021-12-09
Update Microsoft.Azure.Services.AppAuthentication and System.Data.SqlClient packages

Havunen:
Fixes https://github.com/DbUp/DbUp/issues/594


sungam3r:
@Havunen Thanks, `master` already has newer version of these dependencies.

PR #605, opened 2022-02-16
EnsureDatabase.For: Fixed issue when postres database name is always 'postgres'

Homainc:
In dbup-postgres value with `Database` key  from connection string is always ignored and replaced by `postgres`, so it isn't possible to set database name via connection string.


AdrianJSClark:
Thanks for this contribution!

PR #608, opened 2022-03-24
Make DoesTableExist virtual

### LINKED ISSUE #607: Make DoesTableExist virtual
I'm attempting to add manticore support, but while manticore supports most mysql-ish syntax, it does not have the concept of a scalar result. If I could override TableJournal.DoesTableExist I think I'd be almost able to do this.

ithielnor:
Fixes #607


programcsharp:
Looks like this should work, but AppVeyor had an issue perhaps while GH was having an issue. Is there a way to re-trigger the build check, or another commit?

ithielnor:
I've merged in the master branch. This is a miniscule change that would enhance the expandability of the core.

sungam3r (approved):

Issue #612, opened 2022-04-14
4.6.0 release


sungam3r:
The last v4.5.0 was published 20.11.2020. 12 commits on `master` since then including bug fixes. @AdrianJSClark @droyad please publish fresh 4.6.0 release.

Also I want to remind you about #404. I'm still ready to help you to maintain this repo - code review/merging/CI stuff/publishing releases/answering questions. Yes, this project will not be the main one, which I help, but nevertheless the situation can move away from the dead point.

Ekkeir:
Are there any plans on release?
Some of the bug fixes would be beneficial for us.

AdrianJSClark:
Sorry, on my part I know the codebase and am able to merge but I don't think I've got access to make a release happen.

sungam3r:
@AdrianJSClark could you please contact @droyad or other maintainer you know directly?

droyad:
Sorry @AdrianJSClark, I've now given you access. I'll have a review of the commits and make a release this week.

droyad:
I had a look, and unfortunately we have some [breaking changes](https://github.com/DbUp/DbUp/pull/559/files#diff-ca3aece64f9430707ed273367a9968bd2cc9d8d50e67d147cfdd92be9c26097bL34-L55) we'll need to revert first

droyad:
![giphy](https://user-images.githubusercontent.com/1687639/166901510-9cd37761-7c72-4229-8a91-75db771bb942.gif)

PR #618, opened 2022-05-04
Revert some changed made in #559 to maintain binary compatibility

droyad:
The following two APIs were removed:
1. https://github.com/DbUp/DbUp/pull/559/files#diff-57c0a543eea2d12709e1fa053c8c5aff51a96be7e0e6e377254fccafd6b006fbL9
2. https://github.com/DbUp/DbUp/pull/559/files#diff-57c0a543eea2d12709e1fa053c8c5aff51a96be7e0e6e377254fccafd6b006fbL36

The first one is compile-compatible, but may still cause a problem if a newer DLL was used against something that was compiled against an older (which is unlikely though).

However the second one breaks compile compatibility.

This unblocks #612

### DIFF
--- a/src/dbup-sqlserver/SqlConnectionManager.cs
+++ b/src/dbup-sqlserver/SqlConnectionManager.cs
@@ -1,7 +1,11 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Data.SqlClient;
 using DbUp.Engine.Transactions;
 using DbUp.Support;
+#if SUPPORTS_AZURE_AD
+using Microsoft.Azure.Services.AppAuthentication;
+#endif
 
 namespace DbUp.SqlServer
 {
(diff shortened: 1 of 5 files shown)


src/dbup-sqlserver/SqlConnectionManager.cs
```diff
@@ -1,7 +1,9 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Data.SqlClient;
 using DbUp.Engine.Transactions;
 using DbUp.Support;
+using Microsoft.Azure.Services.AppAuthentication;
```
AdrianJSClark, line 6:
```suggestion
#if SUPPORTS_AZURE_AD
using Microsoft.Azure.Services.AppAuthentication;
#endif
```

src/dbup-sqlserver/SqlServerExtensions.cs
```diff
@@ -43,6 +43,23 @@ public static UpgradeEngineBuilder SqlDatabase(this SupportedDatabases supported
...
+    [Obsolete("Use \"AzureSqlDatabaseWithIntegratedSecurity(this SupportedDatabases, string, string)\" if passing \"true\" to \"useAzureSqlIntegratedSecurity\".")]
+    public static UpgradeEngineBuilder SqlDatabase(this SupportedDatabases supported, string connectionString, string schema, bool useAzureSqlIntegratedSecurity)
+    {
+        if (useAzureSqlIntegratedSecurity)
+        {
+            return supported.AzureSqlDatabaseWithIntegratedSecurity(connectionString, schema);
+        }
+
+        return supported.SqlDatabase(new SqlConnectionManager(connectionString), schema);
+    }
```
AdrianJSClark, line 61:
```suggestion
#if SUPPORTS_AZURE_AD
    /// <summary>Creates an upgrader for SQL Server databases.</summary>
    /// <param name="supported">Fluent helper type.</param>
    /// <param name="connectionString">The connection string.</param>
    /// <param name="schema">The SQL schema name to use. Defaults to 'dbo' if <see langword="null" />.</param>
    /// <param name="useAzureSqlIntegratedSecurity">Whether to use Azure SQL Integrated Security</param>
    /// <returns>A builder for a database upgrader designed for SQL Server databases.</returns>
    [Obsolete("Use \"AzureSqlDatabaseWithIntegratedSecurity(this SupportedDatabases, string, string)\" if passing \"true\" to \"useAzureSqlIntegratedSecurity\".")]
    public static UpgradeEngineBuilder SqlDatabase(this SupportedDatabases supported, string connectionString, string schema, bool useAzureSqlIntegratedSecurity)
    {
        if (useAzureSqlIntegratedSecurity)
        {
            return supported.AzureSqlDatabaseWithIntegratedSecurity(connectionString, schema);
        }

        return supported.SqlDatabase(new SqlConnectionManager(connectionString), schema);
    }
#endif
```

I had to fence off this in a local build as well because when `SUPPORTS_AZURE_AD` wasn't set for a platform the  `AzureSqlDatabaseWithIntegratedSecurity` extension method was missing.

AdrianJSClark (approved):

Issue #619, opened 2022-05-05
EnsureDatabase.For.PostgresqlDatabase database does not exists error


ant0nkr:
After upgrading to 4.6.0 release dbUp started to fail all upgrades if database doesn't exists.

because in PostgresqlExtensions.cs on line 163 there is a method being called connection.Open() however in commit https://github.com/DbUp/DbUp/commit/06ea1bde6f9a65b0b48179e90c88c317fac0663d there was database name changed removed. which is caused database creation issue.

to reproduce the issue:  
var adminConnectionString = "Server=localhost;Port=5432;Database=420126721_UnitTestDb;Userid=postgres;Password=12345"
EnsureDatabase.For.PostgresqlDatabase(adminConnectionString);

Make sure 420126721_UnitTestDb doesn't exists.

![image](https://user-images.githubusercontent.com/36110197/166980246-bdd2e4ec-3d65-4442-9bd7-b5cb8a87c36d.png)

Suggestion: 
https://github.com/DbUp/DbUp/pull/620

PR #620, opened 2022-05-05
- fixed issue with database creation for Postgres Database Provider.

ant0nkr:
Fixed an issue described in a ticket: https://github.com/DbUp/DbUp/issues/619

it should be changed, but after all the loggings before actual usage. In this case - logging will be correct and database will be created.

### DIFF
--- a/src/dbup-postgresql/PostgresqlExtensions.cs
+++ b/src/dbup-postgresql/PostgresqlExtensions.cs
@@ -153,6 +153,8 @@ public static class PostgresqlExtensions
 
         logger.WriteInformation("Master ConnectionString => {0}", logMasterConnectionStringBuilder.ConnectionString);
 
+        masterConnectionStringBuilder.Database = "postgres";
+
         using (var connection = new NpgsqlConnection(masterConnectionStringBuilder.ConnectionString))
         {
             if (certificate != null)


droyad:
@Homainc as the author of the original PR #605, does this sound right to you? Could you verify that it doesn't regress your fix.

I can't figure out how this changes anything, the original PR was light on the details of the problem. But I haven't looked closely and am unfamiliar with Postgres.

Homainc:
It seems right to me, I've found out that my problem occured because I don't have access to "postgres" database, which is used to check if specified database is exist. So I can't check if database is exist this way. I will try to find out work around for my case

droyad:
In that case, I think reverting the previous PR makes the most sense. For me it looked like the location in this PR is a bit late and `logMasterConnectionStringBuilder` would be wrong

ant0nkr:
Hi @droyad logMasterConnectionStringBuilder would represent actual database name to which you're connecting to in all scenarios, except creating.
"postgres" database is used to connect to check whether needed database exists.
Reverting or appying this PR would fix the issue.

The only difference would be, what's being displying in Logs:
In case of reverting PR #605 - in logs you will continue to see "postgres".
in case of applying this PR - in logs you will see actual database name specified in ConnectionString.

It's up to you, however from my pov it's better to see actual database name specified in Connection String, because when you see "postgres" in logs it's a bit confusing.

Thanks.

droyad:
The logs though say it's the `Master ConnectionString`. In SqlServer `master` database is the one that stores all the meta data, I assume `postgres` is the Postgres equivalent.

PR #622, opened 2022-05-06
Replaced artifacts with a push to GitHub Packages

droyad:
I can't see a reason why we wouldn't push all packages to GH Packages

Depends on #621


src/dbup-core/dbup-core.csproj
```diff
@@ -15,6 +15,7 @@
     <PackageIcon>dbup-icon.png</PackageIcon>
     <PackageProjectUrl>https://dbup.github.io</PackageProjectUrl>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
+    <RepositoryUrl>https://github.com/DbUp/DbUp</RepositoryUrl>
```
droyad, line 18: This is needed so that GH can marry the package up to the repo it belongs to. Not sure why it does that though.

AdrianJSClark (approved):

PR #623, opened 2022-05-06
Added Google Cloud Spanner Support

marty501:


droyad:
Thank @marty501, in about a week or two the first provider will be split out into it's own repo. I'd like to hold this until that happens so just to make it a little less work.

Also, do you need `net461`? If not, I think we can just target `netstandard2.0` for any new providers.

marty501:
Hi Rob, splitting in two repos sounds good. I can re-submit the PR against
the new repo, if that helps.

As for net461 - I don't really need it. I simply followed the approach used
in the other packages part of the project.

Thanks,
Martin

PR #624, opened 2022-05-08
Improve Package Health

AdrianJSClark:
The changes here make some minor modifications that should improve the quality of the NuGet package output for DbUp and the various providers.

The changes are:

- Combined common properties into a `Directory.Build.props` file so they do not need to be repeated
- Configured for deterministic builds when constructed in CI
- Added Source Link to enable debugging into the code by referencing the sources here on GitHub

~**Note:** This PR depends on the yet-to-be-merged work from both #622 & #621 and if merged before those will include all their changes as well.~ Dependencies merged.

### DIFF
--- a/src/DbUp.sln
+++ b/src/DbUp.sln
@@ -1,7 +1,6 @@
-
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.29230.61
+# Visual Studio Version 17
+VisualStudioVersion = 17.2.32427.441
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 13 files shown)


droyad:
> Combined common properties into a Directory.Build.props file so they do not need to be repeated

Won't we need to undo this when we pull them out into seperate repos? We'll either put them back into the csproj, or copy the `Directory.Build.props` just for the sake of the one csproj file per repo.

droyad:
Sorry, I squash merged :/

AdrianJSClark:
> Sorry, I squash merged which I don't usually do, but there were some garbage commits in there.

😱 You *monster!*

All good, fixed up pretty easily.

> Won't we need to undo this when we pull them out into seperate repos? We'll either put them back into the csproj, or copy the Directory.Build.props just for the sake of the one csproj file per repo.

I've found a `Directory.Build.props` file is handy to have even with only a couple of project files. Happy to reverse it if you think it clutters stuff up.

droyad (approved):

PR #625, opened 2022-05-09
Revert "EnsureDatabase.For: Fixed issue when postres database name is always 'postgres'"

### LINKED ISSUE #619: EnsureDatabase.For.PostgresqlDatabase database does not exists error

droyad:
Reverts #605
Supersedes #620
Resolves #619

### DIFF
--- a/src/dbup-postgresql/PostgresqlExtensions.cs
+++ b/src/dbup-postgresql/PostgresqlExtensions.cs
@@ -145,6 +145,8 @@ public static class PostgresqlExtensions
             throw new InvalidOperationException("The connection string does not specify a database name.");
         }
 
+        masterConnectionStringBuilder.Database = "postgres";
+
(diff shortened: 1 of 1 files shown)


SteveDesmond-ca:
Thanks, just debugged my way to this this morning -- any idea when 4.6.1 will get published?

PR #637, opened 2022-06-18
improve sample for SQLite

flatrick:
1. By putting each type into their own method, the reader can more quickly discern which information is relevant 
1. By explicitly stating which method is being used, we remove the mental load from the reader of needing to figure out which namespace a function belongs to and thereby making it clearer to the user what is actually being used
1. Add an example where `SharedConnection()` is used, which is most likely the most interesting one for new users where SQLite is the primary database type used


droyad:
Thanks, I [picked your commit](https://github.com/DbUp/dbup-sqlite/commit/d2c90141f9b31f15e38510bcb3ad335f48268b0a) over to the new repository.

PR #644, opened 2022-08-13
Fix InvalidCastException on DB existence check (CockhroachDb)

webcoyote:
When checking for the existence of the database, an object is cast to (int?), which works fine for an Int32 object, but crashes on an Int64 object, which is apparently what gets returned when checking CockroachDb, a Postgres-compatible database.

This PR fixes the error by using Convert.Int32 on the object instead of assuming the object type can safely be cast.

Also fixes a duplicate project references error on netstandard1.3

- Remove duplicate project references error
- Fix InvalidCastException on DB existence check


webcoyote:
This is a pretty straightforward change, but I'm glad to make additional updates if desired. Any chance of getting this merged? Thanks!

sungam3r:
Duplicate project references were fixed in default branch. You may rebase.

webcoyote:
> Duplicate project references were fixed in default branch. You may rebase.

Rebased per request

sungam3r:
@droyad @AdrianJSClark could you please review?

sungam3r (approved):

Issue #570, opened 2021-04-06
Use Microsoft.Data.SqlClient


chriswill:
Referencing [this post](https://devblogs.microsoft.com/dotnet/introducing-the-new-microsoftdatasqlclient/).

I want to use a SQL connection string like 
`Server=myserver.database.windows.net,1433;Database=mydatabase;UID=a;Authentication=Active Directory Interactive`

but the app fails with a **Keyword not supported: 'authentication'** exception.  This is because System.Data.SqlClient doesn't understand the new authentication schemes supported in Microsoft.Data.SqlClient that support Sql Azure.

Upgrading to Microsoft.Data.SqlClient is a relatively painless process and would resolve this issue.  Would you accept a PR for this?

chriswill:
I forked the repo and updated the dbup-sqlserver project to use Microsoft.Data.SqlClient and this resolved the issue.

Hooch180:
When will the support for Microsoft.Data.SqlClient be added?

chriswill:
@Hooch180 For one reason or another it seems like PRs are not being accepted here in a timely fashion.  You're welcome to use my fork https://github.com/chriswill/DbUpReboot if you need Microsoft.Data.SqlClient support.

JCarnall:
@chriswill, many thanks, just grabbed your 'DbUp.Reboot' package and it works a charm using the newer connection string format for Azure Sql

TroyWitthoeft:
@chriswill - Good to see someone out there picking up a reboot for DbUp! Thank you.  Let me know if I can help.  Thanks for pointing out the new connection string `authentication` [keywords ](https://docs.microsoft.com/en-us/sql/connect/ado-net/sql/azure-active-directory-authentication?view=sql-server-ver15#setting-azure-active-directory-authentication) in Microsoft.Data.SqlClient.   FWIW, the overload in  #512 is working for me?  I've also got an example of how to manually do something like `Active Directory Default` does in the `TestConnect()` method of AzureDbUp, [here](https://github.com/TroyWitthoeft/AzureDbUp/blob/1e01c4aeae5a07769ba0c9b5ff389a1a1a498241/Program.cs#L185).

chriswill:
@TroyWitthoeft good luck with your project!

LordBenjamin:
+1 for this request - I'm getting an error after upgrading to .NET 6.0 (which has resulted in having to add `TrustServerCertificate=True` for local dev).

[...]

Issue #432, opened 2019-09-12
RFC: Drop Older Framework Support


AdrianJSClark:
I noticed that @droyad is talking about a 5.0 release. In the lead up to that release I wanted to ask if we wanted to use that as an opportunity to leave behind some of our older framework dependencies?

As an example, `dbup-core` supports these frameworks:
```netstandard1.3;net35;net45;netstandard2.0```

My preference would be to simply support `netstandard2.0`. This probably isn't realistic for the wider audience. Considering the currently-supported frameworks, we probably should include .NET Framework 4.5.2 (4.5 and 4.5.1 are out of support) giving this as the target frameworks:
```net452;netstandard2.0```

Obviously this is just one library as an example, there would need to be some thought put in across the board.

AdrianJSClark:
Guidance from Microsoft on this topic for reference [Cross-platform targeting - Open-source library guidance](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/cross-platform-targeting)

justinjstark:
dbup-sqlserver now requires .NET framework 4.6 to get AppAuthentication support. So if you support 4.5 you're going to need an additional target for 4.6 in that project. Why not make the minimum .NET framework 4.6?

AdrianJSClark:
@justinjstark I'd have no problems with going higher personally. I'd like this to be an opportunity to push things as far into the future as possible, without making DbUp unusable for too many people.

Setting support level based on individual DB support requirements, such as your example of SQL Server "AppAuthentication" support, is something which would be easier if each database technology was more separate as discussed in #431. That way `dbup-core` could be the absolute base level of support however the DB providers like `dbup-sqlserver` would be free to decide on a higher minimum if that was warranted to ensure a particular feature.

You closed #431 @droyad, is splitting things up still in the plans?

AdrianJSClark:
I just wanted to leave this blog post here as extra fodder for those following this issue to chew over.

[.NET Core, .NET 5; the exodus of .NET Framework?](https://blog.marcgravell.com/2020/01/net-core-net-5-exodus-of-net-framework.html)

jafin:
Any movement from 2020?

Motivation to move to netstandard2.0   (or at least fork a v5.x to netstandard2.0)

* SQLite (and possibly other) libraries only support netstandard2.0 in recent versions.  There are bugs that get fixed by levelling up.

[...]

PR #655, opened 2022-11-12
Remove Unsupported Target Frameworks

AdrianJSClark:
DbUp has been very good at keeping support for very old frameworks to maintain compatibility. As time has gone on they have become harder & harder to develop with. Many of the underlying frameworks here aren't even presented in the Visual Studio Installer for VS2022.

This change boosts our target frameworks to the currently supported set from Microsoft. I propose that this change be part of a 5.0 release of DbUp and teams which still require legacy, unsupported frameworks use the most recent 4.x releases. While there was a [plan that I'd outlined previously](https://github.com/DbUp/DbUp/issues/432#issuecomment-1131617191) to split the repositories for version 5.0, however I think we've moved too slowly to include that work now.

Before the final 5.0 build is created we should also update from using `System.Data.SqlClient` to using `Microsoft.Data.SqlClient` to track where Microsoft will be developing for SQL Server in the future which is quite a requested change (see #565 #570 #652).

Note that this change also resolves the current issue where the main branch fails to build.


droyad (approved):

src/dbup-core/dbup-core.csproj
```diff
@@ -30,18 +26,18 @@
...
     <PackageReference Include="System.Reflection" Version="4.3.0" />
-    <PackageReference Include="System.Reflection.TypeExtensions" Version="4.4.0" />
+    <PackageReference Include="System.Reflection.TypeExtensions" Version="4.7.0" />
     <PackageReference Include="System.Diagnostics.TraceSource" Version="4.3.0" />
     <PackageReference Include="System.Dynamic.Runtime" Version="4.3.0" />
-    <PackageReference Include="Microsoft.CSharp" Version="4.4.0" />
+    <PackageReference Include="Microsoft.CSharp" Version="4.7.0" />
   </ItemGroup>
 
-  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.3' Or '$(TargetFramework)' == 'netstandard2.0' ">
```
akirayamamoto, line 39: Was `netstandard1.3` intentionally removed from here?

AdrianJSClark: (Sorry, thought "Resolve Conversation" would also post my comment...)

All the `PackageReference` elements included in this section of the project file are also in the `netstandard1.3` specific group just above. The duplication of those references is one of the major things this PR is trying to fix.

PR #661, opened 2022-11-21
Fix DELIMITER handling in MySQL and Oracle

epenet:
If GetCurrentCommandTextFromBuffer is not empty, then DELIMITER parsing should be ignored.

Fixes DbUp/dbup-oracle#2
Fixes DbUp/dbup-mysql#7


epenet:
Split between:
- #790
- https://github.com/DbUp/dbup-mysql/pull/10
- https://github.com/DbUp/dbup-oracle/pull/6

Issue #653, opened 2022-10-30
 dbup.mysql needs to upgrade the corresponding msyql.data version


wangengzheng:
On my mac computer m1, I use dbup.mysql to run it locally, and an error message appears. Of course, if I upgrade the mysql.data version to the latest version, this problem does not recur.

```
System.Collections.Generic.KeyNotFoundException: The given key '28783' was not present in the dictionary.
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   at MySql.Data.MySqlClient.MySqlField.SetFieldEncoding()
   at MySql.Data.MySqlClient.MySqlField.set_CharacterSetIndex(Int32 value)
   at MySql.Data.MySqlClient.NativeDriver.GetColumnData(MySqlField field)
   at MySql.Data.MySqlClient.NativeDriver.GetColumnsData(MySqlField[] columns)
   at MySql.Data.MySqlClient.Driver.GetColumns(Int32 count)
   at MySql.Data.MySqlClient.ResultSet.LoadColumns(Int32 numCols)
   at MySql.Data.MySqlClient.ResultSet..ctor(Driver d, Int32 statementId, Int32 numCols)
   at MySql.Data.MySqlClient.Driver.NextResult(Int32 statementId, Boolean force)
   at MySql.Data.MySqlClient.MySqlDataReader.NextResult()
   at MySql.Data.MySqlClient.MySqlDataReader.Close()
   at MySql.Data.MySqlClient.MySqlDataReader.Dispose(Boolean disposing)
   at MySql.Data.MySqlClient.MySqlDataReader.Dispose()
   at MySql.Data.MySqlClient.MySqlCommand.ExecuteNonQuery()
   at DbUp.Support.ScriptExecutor.ExecuteNonQuery(IDbCommand command)
   at DbUp.Support.ScriptExecutor.<>c__DisplayClass18_1.<Execute>b__1()
   at DbUp.MySql.MySqlScriptExecutor.ExecuteCommandsWithinExceptionHandler(Int32 index, SqlScript script, Action executeCommand)
   at DbUp.Support.ScriptExecutor.<>c__DisplayClass18_0.<Execute>b__0(Func`1 dbCommandFactory)
   at DbUp.Engine.Transactions.NoTransactionStrategy.Execute(Action`1 action)
   at DbUp.Engine.Transactions.DatabaseConnectionManager.ExecuteCommandsWithManagedConnection(Action`1 action)
[...]

PR #664, opened 2022-11-26
Rev Versions around .NET/Github Actions/MySQL

Jtango18:
Minor tweaks to update part of the toolchain. All backwards compatible, but sitting on Latest where possible.

- Build with .NET 7 now that it is GA (6.0 -> 7.0)
- Update SetupDotnet GHA to latest version (1 -> 3)
- Update to support MySQL Data ^8  - this involved conditional dependency tracking for net standard 1.3 as MySQL has dropped that support for the ^8  #series

I believe this also resolves issue #653

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,10 +20,10 @@ jobs:
         with:
           fetch-depth: 0 # all
 
-      - name: Setup .NET 6.0
-        uses: actions/setup-dotnet@v1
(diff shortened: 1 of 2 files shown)


AdrianJSClark (approved):

AdrianJSClark:
Thanks!

Packaged and released: https://www.nuget.org/packages/dbup-mysql/5.0.10

PR #673, opened 2023-01-03
Add better Postgres command splitter

webcoyote:
Use the SQL command parser from Npgsql to correctly split
Postgres SQL code into separate commands, as RegEx will not
work for more complex statements


droyad:
Superceeded by https://github.com/DbUp/dbup-postgresql/pull/19

Issue #404, opened 2019-04-03
Question - Project Support Status


sungam3r:
Hi! We have been using DbUp for quite a while. It so happened that our project required a deeper understanding of the library. Along the way, I had to look at the current status of the project, open issues/pull requests, etc.

Judging by the GitHub Insights graphs, there is an obvious stagnation in the development of the project since November last year. There are many open issues and pull requests even without comments from maintainers.

We are currently exploring the feasibility of continuing to use DbUp, fork it or migrate to other tools. Also there are some things that we would like to add / fix.

So I would like to understand the future development plans. Is this project abandoned or is it just a temporary break? I understand that everyone has his own business and everyone makes his own contribution.

Thanks!

droyad:
@sungam3r Thank you for the interest and gentle inquiry.

I have sounded out my colleges at Octopus Deploy and we are happy to adopt and maintain this project as part of our normal work routine. I and another maintainer @jburger work there and this project originated with the Octopus CEO, Paul. We use this product internally and recommend it to our customers.

I have discussed this with some of the current maintainers, but I still need to get in touch with others. Pending a consensus, we'll make it official.

What will this mean for the project? Other than a gradual increase in momentum on the issues and PRs and the appearance of new faces, nothing else will change much. This org and repo will stay as is, and decisions will still be made in the open in this repo.

I'm happy to answer any questions.

droyad:
Just realized this is issue 404. Fitting.

sungam3r:
Thanks for the answer.

That is, do I understand correctly that in the near future it makes sense to continue to contribute to this repository? I have locally some set of changes related to the correction of errors and typos, and with the new functionality. As I said before, we use DbUp on an ongoing basis and would like to continue using it in our daily work, supporting the development and correction of errors.

droyad:
Go for it, put up some PRs. If it's new functionality and it would require work to put into a PR, perhaps do up an Issue to discuss it before investing time in it.

peterdku:
Any news regarding "making it offical" ?

sungam3r:
> Go for it, put up some PRs.

Unfortunately PRs wait for months without a comment. It is very annoying and demotivates to develop the project further.

Nik-Kudrin:
Any news, guys ?

rain84:
> > Go for it, put up some PRs.
> 
> Unfortunately PRs wait for months without a comment. It is very annoying and demotivates to develop the project further.

It's annoing..
[...]

PR #682, opened 2023-02-18
Allow custom DB creation script to be passed.

jez9999:
Fixes issue DbUp/dbup-sqlserver#16 - allows custom SQL to be passed to create database.


jez9999:
Whilst this does change the public API, it does so in a non-breaking way, returning a bool and adding default params that have identical behaviour to before.

jez9999:
"HttpError: Resource not accessible by integration" - no idea why the build failed.  Doesn't seem to be an issue with the code.

sungam3r:
"HttpError: Resource not accessible by integration" is a known issue. But a more serious problem is the level of support for this project 😢 .

jez9999:
Hmm, I've seen worse.  But it's a bit of a shame, as DbUp is a pretty nice way of updating a DB IMHO.

jez9999:
@AdrianJSClark Are you still around on this project?  How about making some more people maintainers?  It does seem like updates have slowed and I think it'd be a shame if the project died.

sungam3r:
> How about making some more people maintainers? It does seem like updates have slowed and I think it'd be a shame if the project died.

Rel: #404

Justin-Aurizon:
I am guessing maybe the build api key has expired?

sungam3r:
Nope. It's a known issue with test reporter. It should be configured properly to work with PRs from forks.

jez9999:
How are they breaking?

sungam3r:
Binary breaking.

sungam3r:
> a non-breaking way, returning a bool and adding default params that have identical behaviour to before.

Your changes are source compatible i.e one need no action to write some new code **but** code is still needed to be recompiled.

sungam3r:
>  If I'm upgrading a library, I'd expect to need to recompile the code.

No. Remember about transitive dependencies. DbUp is library (or package, whatever). It's consumed by apps or other packages. These packages are consumed by some other packages and so on. Your consumers will be unhappy if their app crashes after updating some intermediate lib from v1.0.0 to v1.0.1 or from v1.0.0 to v.1.1.0 because of signatures' changes in that lib.

jez9999:
That's what Nuget is for, the intermediate packages should mark a new required version of the transitive dependency.

mjauernig:
First at all. Thank you @jez9999 for contributing. 

> @jez9999 no, changes are breaking, I suggest to add `develop` branch for v6 and merge all breaking changes there, ping @mjauernig for comments.

I think we should use [Semantic Versioning 2.0.0](https://semver.org/). 
> Major version X (X.y.z | X > 0) MUST be incremented if any backward incompatible changes are introduced to the public 
 API. It MAY also include minor and patch level changes. Patch and minor versions MUST be reset to 0 when major version is  incremented.
 
 So, i would follow @sungam3r and would say we merge it to v6. Because their are a lot of packages which are build on top of [DbUp.Core](https://www.nuget.org/packages?q=dbup). 
 
 I think after we cleaned the ```pull requests``` and ```issues```. We can create a DbUp Version 6 for the existing ```breaking pull requests``` and release them shortly and then go on to v7.

jez9999:
> First at all. Thank you @jez9999 for contributing.
> 
> > @jez9999 no, changes are breaking, I suggest to add `develop` branch for v6 and merge all breaking changes there, ping @mjauernig for comments.
> 
> I think we should use [Semantic Versioning 2.0.0](https://semver.org/).
> 
> > Major version X (X.y.z | X > 0) MUST be incremented if any backward incompatible changes are introduced to the public
> > API. It MAY also include minor and patch level changes. Patch and minor versions MUST be reset to 0 when major version is  incremented.
> 
> So, i would follow @sungam3r and would say we merge it to v6. Because their are a lot of packages which are build on top of [DbUp.Core](https://www.nuget.org/packages?q=dbup).
> 
> I think after we cleaned the `pull requests` and `issues`. We can create a DbUp Version 6 for the existing `breaking pull requests` and release them shortly and then go on to v7.

I still don't get it.  You're using a package manager like Nuget which specifies your dependency tree.  If something requires a newer version of DbUp, the tree should update and you should get the newer version of DbUp.  Who is dropping in a DLL without using a package manager?

sungam3r:
>  Who is dropping in a DLL without using a package manager?

Forget about "dropping dll". The problem of binary incompatibility is well-known one in nuget-driven world of .NET.

Consider dependency graph:

1) Your app -> fancy_package -> some_other_package -> dbup

2) Your_app -> dbup

In your solution you reference/use DbUp explicitly AND use it implicitly (you may not event know about it) through some dependencies.

One day you notice that you **can** update to dbup v1.0.1 from v1.0.0, so... why not?! You bump version to 1.0.1 and 1.0.1 **should not** break anything according to semver. At least you think so (as I and 1000000 more devs). OK. Now think about your changes again and look at fancy_package or some_other_package. They **were compiled** agains DbUp v 1.0.0 some day. You are not the author of these packages. Now at runtime these packages may call APIs from DbUp that do not exist anymore. BOOM! `MethodNotFoundException` here!

https://learn.microsoft.com/en-us/dotnet/core/compatibility/categories#binary-compatibility

sungam3r:
@mjauernig I removed `version6` label, milestones are used for such things.

jez9999:
Hmm, I guess I'm too used to thinking about it from the npm way of doing things where it just downloads multiple versions of the package when they're required by different things.

droyad:
Picked across as https://github.com/DbUp/dbup-sqlserver/pull/20, will go into 6.0

PR #692, opened 2023-07-15
Fix for issue/bug #691 (making firebird integration work again)

### LINKED ISSUE #691: Firebird integration has degraded
Hi,

I am the original creator of the firebird integration for dbUp 9 years ago. It has unfortunately degraded over time.

1. The JournalTable SchemaVersions table was originaly created using an unquoted tableName. Someone has changed this so it is created as "schemaversion" i.e. quoted. This does not work when a Trigger is later created that references the table unqouted. It is easily fixed and I will provide a PR. Since this was a general change for all databases it indicates to me that Firebird was not tested after the change.

2. WithTransactionPerScript deployments fails when the SchemaVersions table is not created yet. The reason is that the Trigger operates outside the Transaction and tries to upgrade the Id of the SchemaVersions table. Which does not exist. Giving the error: "FirebirdSql.Data.FirebirdClient.FbException (0x80004005): Dynamic SQL Error
SQL error code = -204
Table unknown
schemaversions"

I will also look into fixing nr. 2. Until then the workaround is to do an intial run without WithTransactionPerScript when SchemaVersions has been created it will work afterwards.

Jeern:
Here a some fixes for the firebird provider. I am the original creator of that provider 9 years ago. Unfortunately it has been deteriorating ever since and now it no longer works with Transactions and also it no longer works with the default JournalTable.

This is all fixed in this PR.

As mentioned in the bug the firebird trigger created by `CreateTriggerSql` did not work with or without a transaction. This has been fixed in that method.

Furthermore it is not allowed to Create the journal table and write to it within the same transaction in firebird as explained here:
https://stackoverflow.com/questions/66537195/i-cant-run-inserts-in-firebird-2-5-table-unknown
It fails everytime when a new database with no scripts and no journal tables is used.

This is fixed by using the new bool UseTheSameTransactionForJournalTableAndScripts which is by default true (i.e keeping the normal behaviour) but false for the FirebirdScriptExecutor

The fix works for TransactionPerScript but naturally not when using a single transaction for all scripts. That is why is now disallowed for Firebird by using the new AllowedTransactionModes Flag enum. Default is "All" but for Firebird only TransactionPerScript and no transactions is allowed.

BTW: The fixes have all been tested extensively on a firebird 3.0.10 server

When merged closes #691

### DIFF
--- a/src/dbup-core/Engine/Transactions/AllowedTransactionMode.cs
+++ b/src/dbup-core/Engine/Transactions/AllowedTransactionMode.cs
@@ -1,0 +1,28 @@
+﻿using System;
+
+namespace DbUp.Engine.Transactions
+{
+    /// <summary>
+    /// makes it possible to set up the TransactionModes allowed for a Database provider
+    /// </summary>
+    [Flags]
+    public enum AllowedTransactionMode
+    {
+        None = 0,
+
+        /// <summary>
+        /// Its allowed to run single transactions for an entire run
+        /// </summary>
+        SingleTransaction = 1,
+        /// <summary>
+        /// Its allowed to run transactions pr. script
+        /// </summary>
+        TransactionPerScript = 2,
+        /// <summary>
+        /// Its allowed to run a single transaction that rolls back at the end
+        /// </summary>
+        SingleTransactionAlwaysRollback = 4,
+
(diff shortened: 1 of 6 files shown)


Jeern:
When merged closes #691

mjauernig (approved):

PR #694, opened 2023-08-11
Split CI Workflow in two different Workflows. One for CI and one for release

mjauernig:
On each pull request the _ci_ workflow fails. It looks like it fails of insufficant rights of the user who is creating the pull request.
image
 
 So i split the workflow. One workflow for ci and one workflow for _create draft release_. The draft-release workflow looks for pull requestes on_releases/**_ branch. This has to check, if we use _releases_ branches in the future.

### DIFF
--- a/.github/workflows/create-draft-release.yml
+++ b/.github/workflows/create-draft-release.yml
@@ -1,9 +1,9 @@
-name: Create Draft Release 
+name: Create Draft Release
 
 on:
-  push:
+  pull_request:
     branches:
-      - 'releases/**'
+      - "releases/**"
   workflow_dispatch:
 
 jobs:
@@ -41,9 +41,9 @@ jobs:
         working-directory: artifacts
         run: dotnet nuget push *.nupkg --api-key ${{ env.PACKAGE_KEY }} --source "https://nuget.pkg.github.com/DbUp/index.json"
         env:
-            PACKAGE_KEY: ${{ secrets.GITHUB_TOKEN }}
+          PACKAGE_KEY: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: 'Create Draft Release'
+      - name: "Create Draft Release"
         if: ${{ format('{0}', github.ref_name) == 'master' && format('{0}', env.GITHUB_TOKEN) != '' }}
         shell: pwsh
         working-directory: artifacts
@@ -54,7 +54,7 @@ jobs:
 
       - name: Test Report 🧪
         uses: dorny/test-reporter@v1
-        if: success() || failure()    # run this step even if previous step failed
+        if: success() || failure() # run this step even if previous step failed
         with:
           name: Tests
           path: artifacts/TestResults/*.trx
--- a/.github/workflows/create-draft-release.yml
+++ b/.github/workflows/create-draft-release.yml
@@ -1,0 +1,61 @@
+name: Create Draft Release 
+
+on:
+  push:
+    branches:
+      - 'releases/**'
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: windows-latest
+
+    env:
+      DOTNET_NOLOGO: true
+      DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true # Avoid pre-populating the NuGet package cache
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0 # all
+
+      - name: Setup .NET 7.0
+        uses: actions/setup-dotnet@v3
+        with:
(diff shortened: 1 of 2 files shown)


mjauernig:
This also resolves that every pull request is automatically creating a draft release.

sungam3r:
@mjauernig Regarding CI/CD stratagy I would like to apply here the same approaches as in GraphQL.NET - other project that I work on for a long time. See https://github.com/graphql-dotnet/graphql-dotnet/tree/master/.github/workflows . No releases branches. CI on each PR. CD on tagging. GitHub releases feature actually does tagging for you and autogenerates the list of changes so it's very easy to make a release. If you have time and interest please check workflows in GraphQL.NET repo, I can answer question if any,

mjauernig:
I take a look on the graphql workflows. Gladly we can use an approved workflow strategy.

- CI: test-code.yml 
- CD: I found no yml on graphql, which is triggered by tags. I modified the create-draft-release.yml to get triggered by push + tag _v*_
- Publishing: publish-release.yml is triggered by published releases

.github/workflows/create-draft-release.yml
```diff
@@ -1,10 +1,9 @@
-name: CI
+name: Create Draft Release
```
sungam3r, line 1: What is _draft_ release? Why is it needed?

sungam3r: Oops, did not read to the end of file. OK, I got it, but let's publish preview packages **each** commit to `master`. It's so simple and does not require tagging.

.github/workflows/test-code.yml
```diff
@@ -0,0 +1,45 @@
...
+      - master
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: windows-latest
+
+    env:
+      DOTNET_NOLOGO: true
+      DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true # Avoid pre-populating the NuGet package cache
```
sungam3r, line 15:
```suggestion
      DOTNET_CLI_TELEMETRY_OPTOUT: true
      DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true # Avoid pre-populating the NuGet package cache
```

sungam3r:
> I found no (template) yml on graphql, which is triggered by tags.

Because it's not needed. Each commit into master ends with publishing all packages into GitHub registry. Simple.

.github/workflows/create-draft-release.yml
mjauernig: I removed the creation of the preview packages, because after i merged a few pr. Each time a _release_ where created with an incremented version number. So i thought, it would be better to remove this step. Because either we have to change the version number on release or we have gap in version schema. 

Maybee their is a solution with creating a draft release without version number incrementing?

image

sungam3r: I need to dive into CI scripts here. I don't know yet.

mjauernig:
I added the original _Create Draft Release_ to test-code.yml again. 

I think we should approve the commit to get the ci build working.
image

The publishing workflow we can look later on.

mjauernig:
@sungam3r 
What is your recommendation for branch updating in pull requests?
image

Update with merge or rebase?

sungam3r:
I tend to use `update branch` feature to test my changes with actual `master` if I see that button above. I use `merge` option bacause it is more simple. In fact it does not matter because any option works for **your** branch only. Then when approved your branch will be eventually merged into master and in case of using `squash and merge` merging strategy commits history will be clean and linear.

sungam3r:
I'll review this one at weekend.

sungam3r:
One more consideration - you can safely ignore suggestion to update your branch with changes from base branch if you are sure that changes in base branch do not affect your work in any way. For example, readme was changed in base branch. In that case updating will only slow down your work because of CI round-trip.

mjauernig:
Thanks for the info for _update branch_.

.github/workflows/test-code.yml
```diff
@@ -0,0 +1,55 @@
+name: Test Code
+
+on:
+  pull_request:
```
sungam3r, line 4: Why didn't you add `push: branches: - master` as well?

sungam3r: @mjauernig Regardless of the answer, feel free to merge and let's move on.

sungam3r (approved):

.github/workflows/test-code.yml
sungam3r: I added develop branch for breaking changes. Please add it to scripts where appropriate.

PR #712, opened 2023-08-21
Add badges for issues into README

sungam3r:
@mjauernig I looked through all issues and PRs (some more, some less) and marked them with appropriate labels (not all). I really have little free time for OSS now, but I promised to help with this project. Then I think I can help with review, but do not expect great activity here from me.


mjauernig:
@sungam3r Thank you for your commitment. My time is also limited. But maybe we can push toghether the project back into a normal state. After that we'll see.

PR #718, opened 2023-08-22
Fix for failed build script on github actions

mjauernig:
At the moment every ```CI``` fails because of ```ApprovedFileNotFoundException```

image

```Assent``` try ```_``` instead of ```D:\a\DbUp\DbUp```

If i run the cake script local the files were found. So i guess that in the ```NoPublicApiChangesBase.cs``` the ```UsingName``` ```m.FilePath``` returns a wrong path. The existing ```callerFilePath``` resolves to the same path on local execution. 

So i would like to commit this simple fix to master just for checking. Maybe we are lucky and the right path get resolves. 

If it is not working, i will revert it and setup an private test repository where i can test the build cake and github actions. However, I'm a bit afraid of the effort. 

P.S.
If somebody has more experience how to test and debug github actions, recommendations are welcome.


mjauernig:
The ```basepath``` i commited accidentally. So last commit is to revert it.

jafin:
@mjauernig You can take a look here https://github.com/DbUp/DbUp/pull/721.   I swapped over the assent lib for verifytests, the pathing was simpler, and the assertion tests are passing, but there appears another build error (I didn't get this on my local branch)  Anyhow maybe it helps.

mjauernig:
@jafin Thank you very much. I will check the pr the next days.

Havagan:
> the other [issue with the postgreSQL tests intermittently failing verify](https://github.com/DbUp/DbUp/pull/721/commits/ab1d330d1a191631698cf3aefb00e39dcd146bb5)

I experienced this on #720 and believe it's because there's a shared logger `CaptureLogsLogger` that's passed around through the "RecordingDb" classes and that logger isn't thread safe. Running the tests in parallel in Visual Studio with "Run until Failure" exposes the issue consistently.

After [updating the logger to be thread safe](https://github.com/DbUp/DbUp/pull/724/commits/b8f0389206cbccd274c7055536cf567563c7a1fa), the errors seem to be resolved.

mjauernig:
@jafin Thank you. The replacement of Assent with Verify resolved the failed build. 

Closing this pull request. Continuation for the now failed `Create Draft Release` at #726

PR #726, opened 2023-08-29
Split Workflows

mjauernig:
Thanks to @jafin pr #721 the `cake build` is working again.

Now it fails on `Create Draft Release`. 
image

It has no permissions for `packages`. I think we can fix this with adding `packages: write` to section `permissions`. See  [Permissions for the GITHUB_TOKEN](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) and the screenshot.

But i prefer to remove temporary the _create draft release_ from the `Test Code Workflow` because every merge of an pull request will increase the version number. I will check (in next weeks) how to modify the `cake build` script that draft version number get reused.

### DIFF
--- a/.github/workflows/create-draft-release.yml
+++ b/.github/workflows/create-draft-release.yml
@@ -1,5 +1,8 @@
 name: Create Draft Release
 
+# The use of this workflow is to trigger it temporary.
+# In a later step this workflow should be merged back to workflow -Test Code-.
+
 on:
   push:
     tags:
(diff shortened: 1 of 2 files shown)


mjauernig:
Without the `Create Draft Release` the check were successful.

[https://github.com/DbUp/DbUp/actions/runs/6016434368/job/16320424445](https://github.com/DbUp/DbUp/actions/runs/6016434368/job/16320424445)

Issue #709, opened 2023-08-21
Remove DbUp.Tests.Common from packages?


sungam3r:
What is it?
![изображение](https://github.com/DbUp/DbUp/assets/21261007/68399b59-6d68-47c1-b1a0-e37c959f95aa)

Havagan:
Is this as simple as adding `<GeneratePackageOnBuild>true</GeneratePackageOnBuild>` to the DbUp.Tests.Common.csproj?

Havagan:
PR https://github.com/DbUp/DbUp/pull/731

You can assign this issue to me if you like.

mjauernig:
I will gladly assign it to you. Thank you very much.

droyad:
This package is the shared testing components used by the provider libraries, we only publish it to the GitHub package repository.

PR #731, opened 2023-09-04
Issue #709 - Add project attributes to exclude from nuget pack.

Havagan:
Tested with Cake build and a .nupkg file was not generated for the Tests.Common project.

Closes issue #709.


mjauernig (approved):

droyad:
This package is the shared testing components used by the provider libraries, we only publish it to the GitHub package repository.

Issue #734, opened 2023-09-16
Suggest: a way to modify the script name in EmbeddedScriptsProvider


osexpert:
I made a branch where I added a delegate in EmbeddedScriptsProvider:
https://github.com/osexpert/DbUp/tree/AddOptionToCustomizeScriptName

With this I can more easily trim the namespace from the embedded resources in own code.

osexpert:
I can do the same today by copy\paste the code in EmbeddedScriptsProvider and put into my own class that implement IScriptProvider. But I don't like this solution, because the code must the in sync with changes done in EmbeddedScriptsProvider.

mjauernig:
Hi, 

thank you for the enhancement suggestion. But i didn't quiet understand, where do you want to do the cusomization? Overwriting the `GetScriptNameFromResourceName`?

osexpert:
"Overwriting the GetScriptNameFromResourceName?"
Yes.

I guess the next step would be to add extension method(s) that can utilize it, eg:
```
    public static UpgradeEngineBuilder WithScriptsEmbeddedInAssembly(this UpgradeEngineBuilder builder, Assembly assembly, Func<string, bool> filter, Func<string, string> scriptNameFromResourceName)
    {
        return WithScripts(builder, new EmbeddedScriptProvider(assembly, filter){ GetScriptNameFromResourceName = scriptNameFromResourceName});
    }
```
My first thought was to add more arguments\extend constructors of EmbeddedScriptProvider, to enable passing Func<string, string> GetScriptNameFromResourceName via ctor(s). But having it as a separate property may be "cleaner"? But I can change it to be a part of the constructor, it you think this is better.

It can also be used it without extension methods, like this:
```
var esp = new EmbeddedScriptProvider(Assembly.GetExecutingAssembly(), resourceName => true /* no filter */){ GetScriptNameFromResourceName = resourceName => GetMyCustomScriptName(resourceName));

builder.WithScripts(esp.GetScripts(null));

```

osexpert:
But possibly it could be taken even further...by having it in the UpgradeConfiguration
```
 public class UpgradeConfiguration
{
 public Func<string, string> GetScriptNameFromResourceName = resouceName => resouceName;
}
```
...and then could do this:
```
builder.WithScriptNameFromResourceName(resourceName => GetMyCustomScriptName(resourceName));
builder.WithScriptsEmbeddedInAssembly(Assembly.GetExecutingAssembly())
```

osexpert:
[...]

Issue #166, opened 2016-02-09
Remove namespace from embedded scripts


ReimerA:
It would be nice with a built in feature for removing namespaces from the embedded scripts, since otherwise later refactoring in code will cause problems with scripts. (The name of the scripts stored in the SchemaVersions table, are prefixed with namespace, since thats how they are stored as resources in the assembly)

kennethscott:
I'd actually like to take this one step further and allow the user to configure how they want the filename stored.  Perhaps 3 options:

1. Fully qualified names - ex. Project.Folder.Script.sql
2. No project name - ex. Folder.Script.sql
3. Filenames only - ex. Script.sql

I actually just implemented # 3 today.  My argument is we use a forced naming scheme for script files to ensure uniqueness regardless of folder (_CCYYMMDDhhmmss_Description.sql_), so the namespace and folders are unnecessary for us.  In fact, our intent is to allow the scripts to be moved around/rearranged in the folder hierarchy periodically (group into release folders, move to archive folders, etc.) and storing the fully qualified names is obviously a problem if you move a script.

The way I implemented it was honestly not quite as clean as I'd hoped, but it's not bad either.  In a nutshell, I created my own version of **EmbeddedScriptsProvider** and in the **GetScripts** method, I changed it from calling **SqlScript.FromStream()** to my own version of **FromStream** I just stuck at the bottom of my class:

```cs
 public IEnumerable<SqlScript> GetScripts(IConnectionManager connectionManager)
 {
          ........
          // notice the change from SqlScript.FromStream to this.FromStream
           .SelectMany(x => x.ResourceNames.Select(resourceName => this.FromStream(resourceName, x.Assembly.GetManifestResourceStream(resourceName), this.encoding)))
          .......
 }

 private SqlScript FromStream(string scriptName, Stream stream, Encoding encoding)
 {
     using (var resourceStreamReader = new StreamReader(stream, encoding, true))
     {
         string c = resourceStreamReader.ReadToEnd();
         // Drop the project name and any folder names by taking only the last two qualifiers of the scriptname (filename + sql)
         string[] pieces = scriptName.Split('.');
         return new SqlScript(String.Join(".", pieces.Skip(pieces.Count() - 2)), c);
     }
 }
```

Now that I have an IScriptProvider that will do what I want, I simply added a new extension method to use it:

```cs
[...]

PR #735, opened 2023-09-28
Add FAQ: Remove namespace from embedded scripts

mjauernig:
If you read the discussion on #166, you will see that there are a lot of pitfalls. So I suggest that DbUp officially supports only filenames with namespaces.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -12,4 +12,6 @@ We do not recommend this. It has some pitfalls. See the linked source. But if yo
   </ItemGroup>
 ```
 
+_We recommend using it with namespaces only. DbUp is a community project with limited support. So we will not support you, if you run into problems._
+
 Source: [https://github.com/DbUp/DbUp/issues/166](https://github.com/DbUp/DbUp/issues/166)
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -1,1 +1,15 @@
-Please contribute some FAQ's!
+# FAQ
+
+## How to remove namespace from embedded scripts?
+
+We do not recommend this. It has some pitfalls. See the linked source. But if you really want to. Add to the csproj file:
+
+```
+ <ItemGroup>
+    <EmbeddedResource Include="Database\Migrations\*.sql">
+      <LogicalName>%(filename)%(extension)</LogicalName>
+    </EmbeddedResource>
+  </ItemGroup>
+```
+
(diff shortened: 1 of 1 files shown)


osexpert:
I have a counter suggestion: https://github.com/DbUp/DbUp/pull/737
IMO It is better to give a supported way to do it instead of saying "not supported". I (and others) will be doing it anyways, its just a little bit harder to do it now, when there is no supported way to do it (eg. manually "cloning" and giving the scripts new names, like I currently do).

mjauernig:
The problem is that none of the creators of DbUp are active, so I don't know which solution they prefer. Also, there is no manpower to support people if they run into problems when changing the script name. That's why I propose not to support this feature, and to close #734 and issue #166.

osexpert:
Well, I propose the opposite, and here we are...

mjauernig:
> What matters if what the current\active\voting developers means.
I think the problem is that there is really no strategy for the project, and therefore no guidelines as to what changes I can or cannot add. 

Can you change your counter proposal #737 so that it works like an extension like [DbUpX](https://github.com/fiscaltec/DbUpX)? I think that would be the best solution. I can keep DbUp slim and you can add the functionality without "cloning code". And if someone else needs the functionality, they can install it via nuget.

osexpert:
"Can you change your counter proposal https://github.com/DbUp/DbUp/pull/737 so that it works like an extension like [DbUpX](https://github.com/fiscaltec/DbUpX)?"
I think it belong in DbUp itself. It can actually make DbUp slimmer in the long run, by replacing many of the existing extension methods. If not getting it into DbUIp itself, I don't think it is that valuable (it can be done manually with some extra code).

osexpert:
I see that ScriptInstanceProvider also has a concept of a "namer". So someone has probably thought about this before. So it seems there are several ways to "rename" scripts, either via WithScript\WithScripts that has this "namer" or as I currently do, simple rename them myself via cloning. Both of these solutions suffer from being not very user friendly to do in what seems to be the most common way: WithScriptsEmbeddedInAssemblies.

WithScriptsFromFileSystem also lack this, and has similar problem as WithScriptsEmbeddedInAssemblies:
too many extension methods (product of all combinations of all arguments). So this would be my next "victim":
Create a new WithScriptsFromFileSystem extension method with a configure action, that could replace all other extension metods, and support a "namer" directly.

PR #737, opened 2023-09-30
Add option to modify embedded script names via ScriptNameProvider

osexpert:
Solve #166 and #734 

Can use like:

```
// This works if the scrips has a file ending, eg. MyScript.sql
private static string GetFileNameFromResourceName(string resourceName)
        => string.Join('.', resourceName.Split('.').TakeLast(2));

builder.WithScriptsEmbeddedInAssemblies(Assembly.GetExecutingAssembly(), configure => 
{
 configure.ScriptNameProvider = resourceName => GetFileNameFromResourceName(resourceName);
});

```
This extension method follow the same configure logic seen in AspNet, and with this, only one WithScriptsEmbeddedInAssemblies method is needed and is easily extendable with extending EmbeddedScriptsOptions instead of adding new extension methods.

### DIFF
--- a/src/dbup-core/ScriptProviders/EmbeddedScriptsProvider.cs
+++ b/src/dbup-core/ScriptProviders/EmbeddedScriptsProvider.cs
@@ -16,6 +16,7 @@
         readonly Assembly[] assemblies;
         readonly Encoding encoding;
         readonly Func<string, bool> filter;
+        readonly Func<string, string> scriptNameProvider;
         readonly SqlScriptOptions sqlScriptOptions;
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


osexpert:
So if this PR get approved, I will suggest to remove all\most other WithScriptsEmbeddedInAssemblies extension methods in the next breaking version.

osexpert:
My initial idea was to make configure optional (configure = null) but then the extension methods may collide with these two:
```
public static UpgradeEngineBuilder WithScriptsEmbeddedInAssemblies(this UpgradeEngineBuilder builder, Assembly[] assemblies)
public static UpgradeEngineBuilder WithScriptsEmbeddedInAssembly(this UpgradeEngineBuilder builder, Assembly assembly)
```
But in the next breaking version, configure can be made optional (configure = null) and then remove the other two.

src/dbup-core/ScriptProviders/EmbeddedScriptsProvider.cs
```diff
@@ -36,13 +37,29 @@ public EmbeddedScriptsProvider(Assembly[] assemblies, Func<string, bool> filter,
...
+        public EmbeddedScriptsProvider(Assembly[] assemblies, Func<string, bool> filter, Func<string, string> scriptNameProvider, Encoding encoding, SqlScriptOptions sqlScriptOptions)
         {
             this.assemblies = assemblies ?? throw new ArgumentNullException(nameof(assemblies));
             this.filter = filter ?? throw new ArgumentNullException(nameof(filter));
+            this.scriptNameProvider = scriptNameProvider ?? throw new ArgumentNullException(nameof(scriptNameProvider));
             this.encoding = encoding ?? throw new ArgumentNullException(nameof(encoding));
             this.sqlScriptOptions = sqlScriptOptions;
         }
 
+
```
Zux-Altron, line 61: please remove ws

PR #746, opened 2023-10-19
add repository URL to nuget packages

dhduvall:
I noticed that [renovate](https://github.com/renovatebot/renovate) wasn't setting the source URL or getting the changelogs for any of the `DbUp` packages except for the original obsolete one. Also the source repo link on nuget.org is missing. I think this is the information that's needed to rectify both situations. I couldn't find a way to get this to happen automatically based on the git configuration, so they're all hard-coded. If there's a better way, I'm happy to make the change.

I'll note that I haven't actually tested either issue is fixed by this change, but I'll try to devise an approximation if needed.

### DIFF
--- a/src/DbUp.Console/DbUp.Console.csproj
+++ b/src/DbUp.Console/DbUp.Console.csproj
@@ -14,6 +14,9 @@
     <TargetFrameworkProfile>Client</TargetFrameworkProfile>
     <FileAlignment>512</FileAlignment>
   </PropertyGroup>
+  <PropertyGroup>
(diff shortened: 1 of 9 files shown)


droyad (approved):

dhduvall:
Let me know if there's anything I can do for this. It didn't seem to me that the failing check was something I introduced, but if it's something I need to fix, I may need some help.

droyad:
Thanks. I've [cherry picked](https://github.com/DbUp/DbUp/commit/87d802e45da99eb9e301cd51facb19ed01ef5923) this across to fix the merge errors and to get this closed.

PR #761, opened 2023-11-14
Update index.md

JohnGoldInc:


droyad:
Thanks. I had a look at the doco and I think it's better we remove the VS/VSCode instructions

Issue #720, opened 2023-08-23
Replace LibLog with Microsoft's logging abstractions


Havagan:
With [LibLog deprecated](https://github.com/damianh/LibLog/issues/270), is there a desire to replace it with [Microsoft.Extensions.Logging.Abstractions](https://www.nuget.org/packages/Microsoft.Extensions.Logging.Abstractions)?

mjauernig:
Yes, we should replace it.

Havagan:
@mjauernig I can try and take this.

Havagan:
@mjauernig Logging abstractions only supports netstandard2.0 and up. Does this mean we remove support in dbup for netstandard1.3?

mjauernig:
@Havagan Sure you can grab this. Thank you.

I would say we should drop netstandard1.3. Microsoft also recommend this:
> We recommend you target .NET Standard 2.0, unless you need to support an earlier version. Most general-purpose libraries should not need APIs outside of .NET Standard 2.0. .NET Standard 2.0 is supported by all modern platforms and is the recommended way to support multiple platforms with one target.

Source: [.net Standard Microsoft](https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0)

Another reason for me is: As a community project with limited ressources we can not support any old framework/standard.

@sungam3r Do you agree?

Because this is breaking we would merge the change to release 6 or 7.

Havagan:
Updated tests to match new logging signatures. All tests pass locally.

![image](https://github.com/DbUp/DbUp/assets/2288850/d62a19bc-6a07-4daf-9a07-17e7300ff02c)

twenzel:
@Havagan any progress on this issue?

Havagan:
@twenzel LibLog has been replaced with logging abstractions and PR #720 created. Since that PR, the tests have been refactored under PR #721 which has caused some merge conflicts that need to be resolved.

Since this issue is a `breaking` change, I didn't prioritize fixing the merge conflicts (because I started a new job three weeks ago). I should have some time this weekend to resolve the merge conflicts and update the PR.

Havagan:
I merged `master` into `feature/ilogger-support` and resolved the merge conflicts. All tests pass after updating the verified snapshot test files to match the new logger signatures.

mjauernig:
Thanks @Havagan for the contribution. I will merge it into the v6 release in the next few weeks.

droyad:
Superceeded by #788

PR #774, opened 2024-01-31
Drop support for `netstandard1.3`

droyad:
This drops support for Netstandard 1.3 as it's not required by any currently supported .NET version.

This will allow us to replace LibLog with [Microsoft's Logging abstractions](https://github.com/DbUp/DbUp/issues/720).

It will also allow us to drop `System.Data.SqlClient`.

I'll put up other Issues/PRs to discuss `net462` and `netstandard2.0/net6` support.


AdrianJSClark (commented):
I appreciate you would like my input, however I don't have the context to make a super-informed decision on this.

My bias would be to push support as far into the future as possible. I'd probably go with `net472;net6.0;` but that might have an impact on some Xamarin or Mono support which I don't have exposure to (also technically I think .NET Framework 4.6.2 is still supported).

AdrianJSClark (approved):

AdrianJSClark:
Re-read my comment and realised I was essentially approving it by saying "I'd take the most recent..." 😂

PR #795, opened 2024-06-21
fix: change default character delimiter for Variable Substitution

### LINKED ISSUE #794: The default character for Variable Substitution clashes with a common PostgreSQL idiom
> It is possible to nest dollar-quoted string constants by choosing different tags at each nesting level. This is most commonly used in writing function definitions. For example:

```sql 
$function$
BEGIN
    RETURN ($1 ~ $q$[\t\r\n\v\\]$q$);
END;
$function$
```

[source](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-DOLLAR-QUOTING)

```sql
Create or replace FUNCTION fff(p1 int)
  returns void
  LANGUAGE  plpgsql
AS
$$ --<< outer level quote
DECLARE
 v_Qry  VARCHAR(4000);
BEGIN
  v_Qry := format(
$string$ --<< quote for the string constant passed to the format function
    Create or replace FUNCTION fff_DYNAMIC_SQL()
       returns void
       LANGUAGE  plpgsql
    AS 
    $f1$ --<< quoting inside the actual function body
    DECLARE
      v1  INTEGER;
    begin
      v1 := %s;
      RETURN;
    END; 
    $f1$
$string$, p1);
  EXECUTE v_Qry;
  RETURN;
END; 
$$;
```
[source](https://stackoverflow.com/a/48689807)

ebellani:
closes #794 

This is a backwards incompatible change and would require a major version bump, imho.

### DIFF
--- a/src/dbup-core/Engine/Preprocessors/VariableSubstitutionSqlParser.cs
+++ b/src/dbup-core/Engine/Preprocessors/VariableSubstitutionSqlParser.cs
@@ -26,7 +26,7 @@ public class VariableSubstitutionSqlParser : SqlParser
     /// Delimiter character for variables.
     /// Defaults to `$` but can be overridden in derived classes.
     /// </summary>
-    protected virtual char VariableDelimiter => '$';
+    protected virtual char VariableDelimiter => '!';
 
     /// <summary>
     /// Replaces variables in the parsed SQL
--- a/docs/more-info/variable-substitution.md
+++ b/docs/more-info/variable-substitution.md
@@ -21,12 +21,12 @@ DeployChanges.To
   .WithVariables(myVariables)
 ```
 
-Then in your database script, you need to enclose all variables in ```$``` signs, for example ```$variable$```:
+Then in your database script, you need to enclose all variables in ```!``` signs, for example ```!variable!```:
 
 ```
--- $TestVariable$ $AnotherVariable$
-print '$TestVariable$'
-SELECT * FROM dbo.$TestVariable$
+-- !TestVariable! !AnotherVariable!
+print '!TestVariable!'
+SELECT * FROM dbo.!TestVariable!
 ```
 
 Will result in:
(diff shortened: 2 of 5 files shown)


gukiboy:
Can you think of a way of making this optional? Like, allowing the user to use one or the other as the delimiter, based on setup. This would not break compatibility.

ebellani:
> Can you think of a way of making this optional? Like, allowing the user to use one or the other as the delimiter, based on setup. This would not break compatibility.

There is already a way of changing this via inheritance. My suggestion with this PR is that it is the wrong default, eaisly breaking a the interaction with a major DBMS

ebellani:
@JakeGinnivan @droyad bump?

fuzzybair:
Any chance of getting this fix? We do not use Variable Substitution with DbUp but not being able to create a schema with Stored Procedures is a problem. https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-DOLLAR-QUOTING

src/dbup-core/Engine/Preprocessors/VariableSubstitutionSqlParser.cs
```diff
@@ -26,7 +26,7 @@ public VariableSubstitutionSqlParser(string sqlText, string delimiter = "GO", bo
     /// Delimiter character for variables.
     /// Defaults to `$` but can be overridden in derived classes.
     /// </summary>
-    protected virtual char VariableDelimiter => '$';
+    protected virtual char VariableDelimiter => '!';
```
fuzzybair, line 29: Will this have a similar issue parsing SQL that uses a C style `!=`  for example `SELECT * FROM task WHERE status != 'Done';`

ebellani: I don't think `!=!`  is a valid SQL statement

droyad:
This is a non-obvious breaking change. I think there is too much code out there that depends on that delimiter. We may consider a PostgreSQL specific solution.

src/dbup-core/Engine/Preprocessors/VariableSubstitutionSqlParser.cs
fuzzybair: as long as there are no spaces allowed that sounds good otherwise given `SELECT * FROM task WHERE status != 'Done' AND enabled != true` may be an issue. sorry for a very contrived example but I was wanting to make sure like "= 'Done' AND enabled " does not get treated as a variable.

PR #804, opened 2024-10-05
Remove unnecessary dependencies

0xced:
Now that DbUp targets .NET Standard 2.0 exclusively it's no longer necessary to have references to the various Microsoft.* and System.* NuGet packages.


droyad:
Also Fixes #806

droyad (approved):

droyad:
Thanks @0xced

Issue #366, opened 2018-09-28
Add ScriptOrderer


lvermeulen:
Scripts are currently always ordered by the script's Name:

In UpgradeEngine.GetScriptsToExecuteInsideOperation:
            var sorted = allScripts.OrderBy(s => s.Name, configuration.ScriptNameComparer);

I want to request the addition of a ScriptOrderer, and am sending a PR to that effect.

droyad:
The PR looks good, just wondering about the use case where you don't want to order by name.

droyad:
Is it to use the data from #370?

droyad:
> just wondering about the use case where you don't want to order by name

I've since come across some use cases:
- When the namespace changes
- Ignore spaces
- You use numbers to order, but run out of numbers

PR #834, opened 2025-09-19
Allow custom sorting for scripts other than by [RunGroupOrder, Name]

Pilchard123:
Much like described in #366, this allows a user to specify an arbitrary sort function for scripts. I know there's `ScriptNameComparer` already, but that only allows changing comparison by name; this would allow sorting by any part or property of the script (or indeed by anything else, if desired). In particular I would use it for sorting scripts topologically/by dependency, though no doubt someone out there has other sort functions they'd want, so I wanted to make this more general than just toposorting.

The current behaviour - sorting by `RunGroupOrder` then by `Name` - is maintained if a user does not specify a different sort function. Specifying a `null` sorting function will cause an exception; use `(s)=>s` for an identity/no-extra-sorting function (which could be a named static, as could the default sort function; thoughts?).

You can do something similar (but not the same) with a `ScriptProvider` that sets `RunGroupOrder` individually for each script, but that doesn't allow for easy sorting across multiple `ScriptProvider`s (in fact, I'm not sure it's even possible in the general case).

I do also have a topological sorter that can be plugged into the sorting function as a `Comparer<SqlScript>` . It isn't in this PR right now (I can add it or make a second PR if it would be appreciated; currently it depends on this PR, but I think I know how I'd rework it to be standalone if necessary) because I didn't want to make the PR too large (though it is only one more class). I know DbUpX does dependency sorting, but DbUpX is archived and when I tried using it a couple of weeks ago it didn't work in modern versions of .NET (or possibly at all; I didn't check that far back). (Ab)using the journal to sort scripts also just feels a bit wrong.


droyad:
@copilot Assess whether this is a breaking change

droyad:
Thank you. I built on it and put up #834 (I couldn't update this PR for some reason)

Pilchard123:
Sorry, I didn't see this; my GH session had expired without my realising and my email inbox is so full of garbage I rarely look in it unless I'm expecting something. Would you be interested in my tidying up the topological sorter (and reworking it to fit the new style, as necessary) and PRing that over?

